PR #9, opened 2020-07-13
timegraph: uplift to PIXI 5.0

MatthewKhouzam:
remove deprecated fields "autoResize" and "roundedPixels".

May have a small change in behavior.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -1,3 +1,5 @@
+import * as PIXI from "pixi.js"
+
 import { TimeGraphCursor } from "../components/time-graph-cursor";
 import { TimelineChart } from "../time-graph-model";
 import { TimeGraphChartLayer } from "./time-graph-chart-layer";
@@ -7,8 +9,8 @@ import { TimeGraphChart } from "./time-graph-chart";
 export class TimeGraphChartCursors extends TimeGraphChartLayer {
     protected mouseIsDown: boolean;
     protected shiftKeyDown: boolean;
-    protected firstCursor: TimeGraphCursor;
-    protected secondCursor: TimeGraphCursor;
+    protected firstCursor?: TimeGraphCursor;
+    protected secondCursor?: TimeGraphCursor;
     protected color: number = 0x0000ff;
 
     constructor(id: string, protected chartLayer: TimeGraphChart, protected rowController: TimeGraphRowController, style?: { color?: number }) {
@@ -37,7 +39,8 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
         document.addEventListener('keyup', (event: KeyboardEvent) => {
             this.shiftKeyDown = event.shiftKey;
         });
-        this.stage.on('mousedown', (event: PIXI.interaction.InteractionEvent) => {
+        
+        this.stage.on('mousedown', (event: PIXI.InteractionEvent) => {
             this.mouseIsDown = true;
             const mouseX = event.data.global.x;
             const xpos = this.unitController.viewRange.start + (mouseX / this.stateController.zoomFactor);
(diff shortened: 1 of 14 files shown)


MatthewKhouzam:
Please be pedantic, I am a java developer, I would love to learn the right way to do things here.

timeline-chart/src/time-graph-container.ts
```diff
@@ -45,14 +45,19 @@ export class TimeGraphContainer {
...
             antialias: true,
-            roundPixels: false,
             resolution: ratio,
-            autoResize: true
         });
         this.application.stage.height = config.height;
 
         this.stage = this.application.stage;
         this._canvas = this.application.view;
+        let app: PIXI.Application;
```
tahini, line 54: If you don't plan to assign this variable again, const is preferred to let. Also, you can put the assignation on the same line and you may not need to  put the PIXI.Application (with proper eslint-typescript configuration, it will tell you if you do or don't).

timeline-chart/src/time-graph-container.ts
```diff
@@ -45,14 +45,19 @@ export class TimeGraphContainer {
...
+        app = this.application;
+        canvas.onresize = function (event) {
+            app.renderer.view.width = canvas.width;
+            app.renderer.view.height = canvas.height;
+            app.render();
+        }
```
tahini, line 60: I think if you do }.bind(this) here, then you can use this.application.renderer, etc. instead of passing by a local variable.

The bind method binds the object in parameter to the "this" in the function.

mad-season:
About a replacement for autoResize: true. Try autoDensity: true

timeline-chart/src/time-graph-container.ts
MatthewKhouzam: removed!

timeline-chart/src/time-graph-container.ts
MatthewKhouzam: removed too!

MatthewKhouzam:
> What's the expected change in behavior? How should I test this?

To compile, from the root run `yarn` then run `yarn start` and look at 0.0.0.0:8080

Next patch, adds this to the readme.

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -7,8 +9,8 @@ import { TimeGraphChart } from "./time-graph-chart";
 export class TimeGraphChartCursors extends TimeGraphChartLayer {
     protected mouseIsDown: boolean;
     protected shiftKeyDown: boolean;
-    protected firstCursor: TimeGraphCursor;
-    protected secondCursor: TimeGraphCursor;
+    protected firstCursor?: TimeGraphCursor;
```
tahini, line 12: Why is this now become optional? Maybe that's the source of the problem?

timeline-chart/src/time-graph-container.ts
```diff
@@ -59,13 +58,13 @@ export class TimeGraphContainer {
         this.layers = [];
 
         const background = new TimeGraphRectangle({
-            opacity: 0,
```
MatthewKhouzam, line 62: THIS WAS THE PROBLEM! transparent according to PIXI5 is incorporeal. So according to PIXI 5 I can fly through glass.

timeline-chart/src/time-graph-container.ts
```diff
@@ -59,13 +58,13 @@ export class TimeGraphContainer {
...
+            opacity: 1,
             position: {
                 x:0, y:0
             },
             height: this.canvas.height,
             width: this.canvas.width,
-            color: 0x550000
+            color: config.backgroundColor,
```
MatthewKhouzam, line 67: since it's not transparent, this is the bg color

MatthewKhouzam:
Pic to show selection working.
select

MatthewKhouzam:
Thanks all!

PR #16, opened 2020-07-28
Improve vertical line rendering performance by eliminating no-op loop…

mad-season:
… iterations


mad-season:
> This is great, trace compass does it too, but we need a signed-off-by: for the patch, see other commit messages for examples.

Done chief.

mad-season:
I don't have write access, so I am relying on a maintainer to merge the request. Thanks much!

PR #18, opened 2020-07-30
Render state labels.

mad-season:
Adaptive font size and color determination.

Signed-off-by: Akshay Sharma <akshay.c.sharma@ericsson.com>

### DIFF
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -47,7 +47,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
 
         if (this._model.label) {
             const fontName = TimeGraphRowElement.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2);
-            const labelTextObj = new PIXI.BitmapText(this._model.label, { fontName: fontName ? fontName : "LabelFont8White" });
+            const labelTextObj = new PIXI.BitmapText(this._model.label, { fontName: fontName ? fontName : TimeGraphRowElement.fontController.getDefaultFontName() });
             this.textWidth = labelTextObj.getLocalBounds().width;
         }
     }
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -1,6 +1,8 @@
 import { TimeGraphComponent, TimeGraphStyledRect, TimeGraphElementPosition } from "./time-graph-component";
 import { TimeGraphRow } from "./time-graph-row";
 import { TimelineChart } from "../time-graph-model";
+import { FontController } from "../time-graph-font-controller"
+import * as PIXI from "pixi.js";
 
 export interface TimeGraphRowElementStyle {
     color?: number
@@ -13,6 +15,8 @@ export class TimeGraphRowElement extends TimeGraphComponent {
 
     height: number;
     position: TimeGraphElementPosition;
+    textWidth: number = 0;
+    static fontController: FontController = new FontController();
 
     protected _options: TimeGraphStyledRect;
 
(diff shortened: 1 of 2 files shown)


timeline-chart/src/time-graph-font-controller.ts
```diff
@@ -0,0 +1,60 @@
...
+    private fontFamily: string;
+    private fontNameMap: Map<number, Map<string, string>>;
+    private fontColorMap: Map<number, string>;
+
+    constructor(fontFamily: string = "\"Lucida Console\", Monaco, monospace") {
+        this.fontFamily = fontFamily;
+        this.fontNameMap = new Map<number, Map<string, string>>();
+        this.fontColorMap = new Map<number, string>();
+
+        PIXI.BitmapFont.from("LabelFont8Black", { fontFamily: this.fontFamily, fontSize: 8, fill: "black", fontWeight: 900 }, { chars: PIXI.BitmapFont.ASCII });
```
MatthewKhouzam, line 13: This is a question, should a function "Name"-> regex like \D+(\d+)(\D+) -> {fontFamily : this.fontFamily, fontSize: $1, fill: $2, fontWeight: 900} be interesting? I see the function being called 4 times, maybe this can avoid copy-paste mistakes?

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -40,6 +44,60 @@ export class TimeGraphRowElement extends TimeGraphComponent {
             borderWidth: _style.borderWidth || 0,
             borderColor: _style.borderColor || 0x000000
         };
+
+        if (this._model.label) {
+            const fontName = TimeGraphRowElement.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2);
+            const labelTextObj = new PIXI.BitmapText(this._model.label, { fontName: fontName ? fontName : "LabelFont8White" });
```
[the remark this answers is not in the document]
mad-season: yep. will do.

timeline-chart/src/time-graph-font-controller.ts
mad-season: you're right. i'll move font creation to a function e.g. createBitmapFont ( \D(\d+)(\D+) ) {...}

MatthewKhouzam: I realized, we should just have the left hand side and generate the label from that. It would be so much simpler.

mad-season: i didn't quite get that...

MatthewKhouzam: We have all the data, why enter it in a concatenated string to split it up again instead of giving the raw data and generating the string out of that.

mad-season: ok so just to be sure, you're referring to the label and not the font right?

timeline-chart/src/time-graph-font-controller.ts
```diff
@@ -0,0 +1,60 @@
...
+                fontColor = "black";
+            }
+            this.fontColorMap.set(color, fontColor);
+        }
+
+        let fontName: string | undefined;
+        if (height) {
+            if (!this.fontNameMap.has(height)) {
+                const blackFont = "LabelFont" + height.toString() + "Black";
+                const whiteFont = "LabelFont" + height.toString() + "White";
```
MatthewKhouzam, line 46: like here

timeline-chart/src/time-graph-font-controller.ts
MatthewKhouzam: I am suggesting a "getFontName" function that takes as arguments the style
so it would be ("LabelFont" + style.fontSize + style.fill)
extra points if fill is capitalized, but as long as it's consistent, I'm happy.

PR #22, opened 2020-08-17
SM: Update chart after collapsing or expanding entries in the tree

smouss:
Added a new condition in time-graph-chart.ts that allows to fetch new
data when collapsing or expanding entries. The function updateChart()
allows to update the chart's rows simulteanously when it is called
in the timegraph-output-component.ts in theia-trace-extension.

Signed-off-by: Soukaina Moussaoui <soukaina.moussaoui@polymtl.ca>

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -36,6 +36,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     protected providedResolution: number;
 
     protected fetching: boolean;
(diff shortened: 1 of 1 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -36,6 +36,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     protected providedResolution: number;
 
     protected fetching: boolean;
+    protected allowToUpdateChart: boolean;
```
MatthewKhouzam, line 39: updateChartEnabled?

PR #25, opened 2020-08-19
Implement zoom in/out and horizontal scrolling using WASD

ankusht-work:
-- Added the mousemove event handler to capture x,y position on the canvas.
-- The x,y position will determine the start from where the zoom should trigger.
-- Recalculated the start and end position.
-- Call the viewRange method with the new coordinates on the canvas.
Note: I have tested this on chrome version (75.0.3770.100) and firefox (79.0 64 bit)

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -21,6 +21,13 @@ export interface TimeGraphChartProviders {
     rowStyleProvider?: (row: TimelineChart.TimeGraphRowModel) => TimeGraphRowStyle | undefined
 }
 
+export const keyBoardNavs: Record<string, Array<string>> = {
+    "zoomin": ['w', 'i'],
+    "zoomout": ['s', 'k'],
+    "panleft": ['a', 'j'],
+    "panright": ['d', 'l']
+}
+
 export type TimeGraphRowStyleHook = (row: TimelineChart.TimeGraphRowModel) => TimeGraphRowStyle | undefined;
 
 export class TimeGraphChart extends TimeGraphChartLayer {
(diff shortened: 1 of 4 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -50,38 +51,102 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     protected afterAddToContainer() {
         this.shiftKeyDown = false;
         this.ctrlKeyDown = false;
+
```
MatthewKhouzam, line 54: Add a method: setControls(map<'string', list<keycode>>)
with the values 
'zoomin'
'zoomout'
'panleft'
'panright'

then....

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -50,38 +51,102 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+            mousePositionY = event.offsetY;
+        })
+
         document.addEventListener('keydown', (event: KeyboardEvent) => {
             this.shiftKeyDown = event.shiftKey;
             this.ctrlKeyDown = event.ctrlKey;
+            let keyPressed = event.keyCode;
+
+            if (triggerKeyEvent) {
+                if (keyPressed === keyboardKey.getCode('w') || keyPressed === keyboardKey.getCode('i')) {
```
MatthewKhouzam, line 110: here map.get('zoomin').contains(keyPressed)

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -50,38 +51,102 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                    const zoomPosition = (mousePositionX / this.stateController.zoomFactor);
+                    const deltaLength = -(mousePositionY / this.stateController.zoomFactor);
+                    adjustZoom(zoomPosition, deltaLength);
+
+                } else if (keyPressed === keyboardKey.getCode('s') || keyPressed === keyboardKey.getCode('k')) {
```
MatthewKhouzam, line 115: And so on...?

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -21,6 +21,13 @@ export interface TimeGraphChartProviders {
     rowStyleProvider?: (row: TimelineChart.TimeGraphRowModel) => TimeGraphRowStyle | undefined
 }
 
+export const keyBoardNavs: Record<string, Array<string>> = {
+    "Zoomin": ['w', 'i'],
```
tahini, line 25: You could use constants here, as the capital letter scheme may easily get mixed up by future developers (start by capital and no camelcase). all caps constants would be clearer. Or make them all lowercase to match case scheme of keyboard and mouse events

ankusht-work: Good point. I have made them all lowercase.

MatthewKhouzam (approved):

PR #30, opened 2020-08-25
Update version to 0.1.0

tahini:
Versions 0.0.2 was tagged and done in 2018, then removed, and brought
back, so let's make a first minor release to make sure we don't go back
in versioning times.

Signed-off-by: Geneviève Bastien <gbastien+lttng@versatic.net>


bhufmann:
We should create a release on NPM for 0.1.0

PR #32, opened 2020-08-25
Fetch time graph chart information for the selected row.

muddana-satish:
Added getPosition, getRowVerticalOffset and getRowHeight functions to get row specific information which is utilized by time-graph-output-component.tsx in theia-trace-extension.

This is a dependency PR for the selection of chart components on click of a row in events-table.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -326,6 +326,20 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         });
         return element as TimeGraphRowElement;
     }
+    
+    getRowVerticalOffset(row: TimelineChart.TimeGraphRowModel): number {
+        const rowComponent: TimeGraphRow | undefined = this.rowComponents.get(row);
+        return (rowComponent ? rowComponent.position.y : 0)
+    }
+
+    getRowHeight(row: TimelineChart.TimeGraphRowModel): number {
+        const rowComponent: TimeGraphRow | undefined = this.rowComponents.get(row);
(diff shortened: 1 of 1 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -326,6 +326,20 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         });
         return element as TimeGraphRowElement;
     }
+    
+    getRowVerticalOffset(row: TimelineChart.TimeGraphRowModel): number {
+        const rowComponent: TimeGraphRow | undefined = this.rowComponents.get(row);
+        return (rowComponent ? rowComponent.position.y : 0)
```
mirsky-work, line 332: Can we handle an undefined `rowComponent` better, since I assume `0` can be also a valid value?
The same applies to the method below. Instead of having these two methods, perhaps have a single getter for the TimeGraphRow object?

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -326,6 +326,20 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+    }
+
+    getRowHeight(row: TimelineChart.TimeGraphRowModel): number {
+        const rowComponent: TimeGraphRow | undefined = this.rowComponents.get(row);
+        return (rowComponent ? rowComponent.height : 0);
+    }
+
+    getPosition(): PIXI.ObservablePoint {
```
mirsky-work, line 340: The method's name might be a little too general.

PR #40, opened 2020-09-15
States not visible in the initial zoom level

muddana-satish:
Set the minimum width of a state to 1 so that small states with negligible widths are visible in the initial zoom level

Signed-off-by: Satish Muddana <satish.muddana@ericsson.com>

### DIFF
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -34,7 +34,8 @@ export class TimeGraphRowElement extends TimeGraphComponent {
             x: this.range.start,
             y: this._row.position.y + ((this.row.height - this.height) / 2)
         };
-        const width = this.range.end - this.range.start;
+        // min width of a state should never be less than 1(for visibility)
+        const width = Math.max(1, this.range.end - this.range.start);
         this._options = {
             color: _style.color,
             height: this.height,
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -34,7 +34,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
             x: this.range.start,
             y: this._row.position.y + ((this.row.height - this.height) / 2)
         };
-        // min width of a state should never be less than 1(for visibility)
+        // min width of a state should never be less than 1 (for visibility)
         const width = Math.max(1, this.range.end - this.range.start);
         this._options = {
             color: _style.color,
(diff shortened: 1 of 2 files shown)


tahini (commented):
This looks good. But how can I reproduce the problem that this fixes?

muddana-satish:
@tahini I couldn't find any trace in 'tracecompass-test-traces' where I can reproduce this issue. I noticed that there are few states that doesn't show up in the time-graph on the initial zoom level though the api returns few states.

The same states starts appearing as you zoom in. This is because the width of those states on the initial zoom level is negligible(almost 0). Hence we need to have a limit on the minimum width to let the user know that there are states present in the chart.

tahini (approved):
ok makes sense

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -34,7 +34,8 @@ export class TimeGraphRowElement extends TimeGraphComponent {
             x: this.range.start,
             y: this._row.position.y + ((this.row.height - this.height) / 2)
         };
-        const width = this.range.end - this.range.start;
+        // min width of a state should never be less than 1(for visibility)
```
PatrickTasse, line 37: Maybe add space after 1

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -263,7 +263,8 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                                 x: this.getPixels(start - this.unitController.viewRange.start),
                                 y: el.position.y
                             },
-                            width: this.getPixels(end - start)
+                            // min width of a state should never be less than (for visibility)
```
PatrickTasse, line 266: less than 1

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -34,7 +34,8 @@ export class TimeGraphRowElement extends TimeGraphComponent {
...
+        const width = Math.max(1, this.range.end - this.range.start);
```
PatrickTasse, line 38: I don't know how this is all used but:
What is range.start and range.end? Are they pixel x-coordinates? If start=4 and end=5, does it mean that state should be drawn from 4 to 5, and in that case does it mean that width should be 2? If all this is correct, I think that the formula should be width = end - start + 1 ?

muddana-satish: @PatrickTasse Yes, range.start and range.end are the x-coordinates adjusted with respect to view range of the chart. These values are from the server.

For your example, I am not sure if we have to set it to '1' or '2' pixels.  But I see for all width calculations, we have used 'end-start'.

timeline-chart/src/components/time-graph-row-element.ts
muddana-satish: Updated.

timeline-chart/src/layer/time-graph-chart.ts
muddana-satish: updated.

MatthewKhouzam (approved):

bhufmann:
I've just tested it as well and it looks good to me too. Thanks for the patch and the reviews!

PR #41, opened 2020-09-18
Use pixi.js-legacy instead of pixi.js

tahini:
This allows support of older browsers without WebGL support.

Signed-off-by: Geneviève Bastien <gbastien+lttng@versatic.net>

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -19,7 +19,6 @@
   },
   "dependencies": {
     "lodash.throttle": "^4.1.1",
-    "pixi.js": "^5.0.0",
     "timeline-chart": "^0.1.0"
   }
 }
--- a/timeline-chart/package.json
+++ b/timeline-chart/package.json
@@ -9,9 +9,10 @@
   },
   "dependencies": {
     "@types/lodash.throttle": "^4.1.4",
+    "glob": "^7.1.6",
     "keyboard-key": "1.1.0",
     "lodash.throttle": "^4.1.1",
-    "pixi.js": "^5.0.0",
+    "pixi.js-legacy": "^5.3.3",
     "rimraf": "latest"
   }
 }
--- a/timeline-chart/src/components/time-graph-arrow.ts
+++ b/timeline-chart/src/components/time-graph-arrow.ts
@@ -1,4 +1,4 @@
-import * as PIXI from "pixi.js"
+import * as PIXI from "pixi.js-legacy"
 
 import { TimeGraphComponent, TimeGraphElementPosition, TimeGraphComponentOptions } from "./time-graph-component";
 
--- a/timeline-chart/src/components/time-graph-axis-scale.ts
+++ b/timeline-chart/src/components/time-graph-axis-scale.ts
@@ -1,4 +1,4 @@
-import * as PIXI from "pixi.js"
+import * as PIXI from "pixi.js-legacy"
 
 import { TimeGraphComponent, TimeGraphInteractionHandler, TimeGraphStyledRect, TimeGraphComponentOptions } from "./time-graph-component";
 import { TimeGraphUnitController } from "../time-graph-unit-controller";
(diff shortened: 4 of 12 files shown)


marcdumais-work (approved):
Changes look good. I compared the subjective performance of the example using Gitpod, between this PR and master. I assume I had the benefit of hw acceleration on master, and I noticed no discernible difference.

MatthewKhouzam (approved):

bhufmann:
> Changes look good. I compared the subjective performance of the example using Gitpod, between this PR and master. I assume I had the benefit of hw acceleration on master, and I noticed no discernible difference.

There should not be a difference on Gitpod. The legacy pixi library still uses HW acceleration as when using pixi.js on master. The only difference is if HW acceleration/WebGL1/2 is not available the legacy library will fallback to canvas because the legacy library includes that code.

marcdumais-work:
> There should not be a difference on Gitpod. The legacy pixi library still uses HW acceleration as when using pixi.js on master. The only difference is if HW acceleration/WebGL1/2 is not available the legacy library will fallback to canvas because the legacy library includes that code.

Ok, thanks for the clarification @bhufmann. Then I think we will need to confirm if it works as hoped, running in a virtualized context with some (inefficient, through CPU) support present for WebGL.

tahini:
One thing I'd like to investigate first is performance. On my laptop, that uses the fallback, when I have the [vscode] trace extension running, my CPU fan runs wild, even though I'm not touching the computer. And that, until I close vscode. I'll need to get in a quiet room and try the current theia browser and electron app and other configuration and listen to the fan, and do more objective performance measurements of course ;-)

ankusht-work:
I will be evaluating this patch it in a virtual environment without GPU and will share my findings.

MatthewKhouzam:
Let's get this in then!

ankusht-work:
**Brief analysis while running it on a VM with no GPU:**

When you force the Pixi to run the canvas renderer either by setting forcecanvas to false or turning off the "Use hardware acceleration when available" setting on chrome settings, there is a significant performance improvement in chart rendering and XY graph responsiveness. Monitored the CPU utilization and it was capped at 70% or less (for one out of the four cores).

When you set the flag to true or allow the hardware acceleration on the browser, the PIXI uses WebGL 1.0 for chart rendering. The CPU utilization goes above 300% and you will see the performance degrading gradually even when you do not perform any action.

MatthewKhouzam:
Oh no! I misread "I have evaluated, not I will be evaluating." But it looks like it seems to still work. :+1:

tahini:
The results of my own investigation is that this patch did not bring performance downgrade. The performance was as bad without!

@ankusht-work iiuc, a simple flag can significantly improve performances? Does this flag have side-effects? If not, then, can you make a patch for it?

ankusht-work:
@tahini : Your patch doesn't improve the performance as you said but it fixes this error "WebGL unsupported in this browser, use "pixi.js-legacy" for fallback canvas2d support" when you force PIXI to use canvas rendering. 
You will see the performance improvement when initializing the PIXI application with canvas renderer vs WebGL (improved performance in case of canvas rendering).


By setting the flag "forcecanvas = true", you will force PIXI to always instantiate the application with canvas rendering (which means no WebGL rendering even if your machine has GPU support). Is that what we want? Or we should make it configurable via UI?. @MatthewKhouzam

MatthewKhouzam:
I would pass a flag, at the least, consider that if you are on a machine with no gl, you can configure it properly from the flag, does that sound sane?

tahini:
I didn't say this patch improves performances, it just doesn't degrade it either and fixes a bug.

From when I looked at the code, I recall that pixi.js-legacy has a method that lets you know if you have GPU support, I'm just not sure we have access to it. But if we do, we could use it to set the flag or not.

bhufmann:
In @ankusht-work case, WebGL1 is used when 'forcecanvas = false' but no GPU and it performs not well. We would have to detect this case somehow to forcecanvas to true ourselves because pixi doesn't do it internally if I'm not mistaken.

ankusht-work:
can we use this method "PIXI.autoDetectRenderer({forceCanvas: true})"?

Found this online via "https://pixijs.download/dev/docs/PIXI.html#autoDetectRenderer"

PR #45, opened 2020-09-27
Change Event listeners from document to Canvas

muddana-satish:

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -66,12 +66,15 @@ const chartHTMLContainer = document.createElement('div');
 chartHTMLContainer.id = 'main_chart';
 container.appendChild(chartHTMLContainer);
 
+const chartCanvas = document.createElement('canvas');
+chartCanvas.tabIndex = 1;
+
 const timeGraphChartContainer = new TimeGraphContainer({
     id: timeGraph.id + '_chart',
     height: styleConfig.mainHeight,
     width: styleConfig.mainWidth,
     backgroundColor: styleConfig.chartBackgroundColor
-}, unitController);
+}, unitController, chartCanvas);
 chartHTMLContainer.appendChild(timeGraphChartContainer.canvas);
 
 const timeGraphChartGridLayer = new TimeGraphChartGrid('timeGraphGrid', rowHeight);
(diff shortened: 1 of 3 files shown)


tahini (changes requested):
I think the problem may be with the fact that the event listeners is added to the document, which, unless I'm mistaken, is the complete web page, and you make no verification of the active widget when you press the key.

For instance, if the focus is in the trace explorer on the left and we press up/down/left/right, both the trace explorer reacts to the key AND the selected state changes!!  I think you should add events to the react widgets that contain the view and add a e.stopPropagation() to the event once you process it.

muddana-satish:
> I think the problem may be with the fact that the event listeners is added to the document, which, unless I'm mistaken, is the complete web page, and you make no verification of the active widget when you press the key.
> 
> For instance, if the focus is in the trace explorer on the left and we press up/down/left/right, both the trace explorer reacts to the key AND the selected state changes!! I think you should add events to the react widgets that contain the view and add a e.stopPropagation() to the event once you process it.

That was my initial thought. But the navigation works even after removing the focus from the chart in eclipse. Having eventlistener on document object ensures the same behavior in trace extension. (I did not want to change this).

tahini:
> But the navigation works even after removing the focus from the chart in eclipse.

What do you mean here? From what I observe, it is true that we can zoom in/out (ctrl + mouse scroll) in the view even if the view is not focused, but the arrows (and key press in general) affect only the view that has focus.

muddana-satish:
@tahini I pushed a commit based on your comments. Please review 0c48fbc8c3dcc491d0e402b835f08ed40b80ba72

tahini (changes requested):
It works really great now! Could you rebase your branch on latest master, as this includes a patch by Avhinava that has been merged since. Also, could you please squash the commits and keep the commit message that best describes what this does now.

muddana-satish:
> It works really great now! Could you rebase your branch on latest master, as this includes a patch by Avhinava that has been merged since. Also, could you please squash the commits and keep the commit message that best describes what this does now.

Squashed all the commits. Could you please verify it now.

tahini (approved):
Works beautfully! Thanks!

MatthewKhouzam (approved):
Great job decoupling it!

PR #48, opened 2020-10-01
Programmatic switch between canvas renderer and WebGL

ankusht-work:
Addresses: https://github.com/theia-ide/theia-trace-extension/issues/122

### DIFF
--- a/timeline-chart/src/time-graph-container.ts
+++ b/timeline-chart/src/time-graph-container.ts
@@ -32,6 +32,8 @@ export class TimeGraphContainer {
         } else {
             canvas = extCanvas;
         }
+
+        const noWebgl2 = !PIXI.utils.isWebGLSupported() || !canvas.getContext('webgl2');
         canvas.style.width = config.width + 'px';
         canvas.style.height = config.height + 'px';
         canvas.width = config.width;
@@ -47,7 +49,8 @@ export class TimeGraphContainer {
             transparent: config.transparent,
             antialias: true,
             resolution: ratio,
-            autoDensity: true
+            autoDensity: true,
+            forceCanvas: noWebgl2
         });
 
         this.stage = this.application.stage;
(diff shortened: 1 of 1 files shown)


bhufmann (approved):
Looks good to me. I've tested both webgl2 and canvas creation by using different environments.

timeline-chart/src/time-graph-container.ts
```diff
@@ -32,6 +32,16 @@ export class TimeGraphContainer {
...
+
+        let hasCanvas = true;
+
+        if (PIXI.utils.isWebGLSupported()) {
+            if (!canvas.getContext('webgl2')) {
+                hasCanvas = true;
+            } else {
+                hasCanvas = false;
+            }
+        }
```
mirsky-work, line 44: IMHO, this form and name makes it slightly more readable.

```suggestion
        let noWebgl2 = !PIXI.utils.isWebGLSupported() || !canvas.getContext('webgl2');
```

tahini:
I don't know too much about PIXI. What are the implications of this? Do we still need pixi.js-legacy? Does this improve the performances on machines without webGL?

ankusht-work:
> I don't know too much about PIXI. What are the implications of this? Do we still need pixi.js-legacy? Does this improve the performances on machines without webGL?

Yes, we still need Pixi.js-legacy. The conclusion is that by switching to canvas renderer on machines with no WebGL support we see overall improved performance.

timeline-chart/src/time-graph-container.ts
ankusht-work: Updated the commit with suggested changes.

tahini (approved):
Code looks good. I can't say I see any performance improvement though, my CPU fans still run like mad without webgl2 support. Is that what you observe too?

tahini:
I just saw the improvement with @ankusht-work 's machine. So I totally approve this PR. My CPU fan's problem is the re-rendering of the canvas, which is totally unrelated to this PR.

bhufmann:
> I just saw the improvement with @ankusht-work 's machine. So I totally approve this PR. My CPU fan's problem is the re-rendering of the canvas, which is totally unrelated to this PR.

Thanks for the approval. Yes, it's probably related to the periodic redraw of the canvas. To not periodic redraw the canvas and redraw on demand will be tracked in a different issue.

PR #49, opened 2020-10-02
Retain row and state selection after refresh

### LINKED ISSUE #50: Zooming in deselects Row in Timegraph chart
After a row and state is selected in timegraph chart, zooming in deselects them. I believe they should be retained.

**Row deselected:**
Row Select Issue

**Actual State deselected:**

State deselects

muddana-satish:
fixes #50 

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -9,7 +9,7 @@ import { TimeGraphChartSelectionRange } from "timeline-chart/lib/layer/time-grap
 import { TimeGraphAxisCursors } from "timeline-chart/lib/layer/time-graph-axis-cursors";
 // import { timeGraph } from "timeline-chart/lib/test-data";
 import { TimelineChart } from "timeline-chart/lib/time-graph-model";
-import { TimeGraphRowElement, TimeGraphRowElementStyle } from "timeline-chart/lib/components/time-graph-row-element";
+import { TimeGraphRowElementStyle } from "timeline-chart/lib/components/time-graph-row-element";
 import { TestDataProvider } from "./test-data-provider";
 import { TimeGraphChartGrid } from "timeline-chart/lib/layer/time-graph-chart-grid";
 import { TimeGraphVerticalScrollbar } from "timeline-chart/lib/layer/time-graph-vertical-scrollbar";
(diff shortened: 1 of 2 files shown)


tahini:
When zooming out, the selection is lost once more. Also, when a state is selected at high zoom level, the selected time may be beside that state when we zoom in. See the screecast here: this is taken with your PR in timeline-chart and theia-trace-extension

rowSelection

muddana-satish:
I think we should test it with actual states. The GAP states doesn't have fixed range and mostly depends on the zoom levels. Could you please test it with actual states.

muddana-satish:
@tahini 
This is how it looks for the actual states. 

State Selection

I believe this functionality is correct. The GAP states keep changing based on the zoom level. So it is highly unlikely to identify the state. Any thoughts on this?

tahini:
I thought I had selected a state at first! I clicked on something, very thin, true, but still, it highlighted the visible state and the state remains hightlighted as I zoom in. This may be unrelated with this PR, but it's a bug anyway. If the state appears to be selected, then I'd expect to zoom in and remain in that highlighted state.

muddana-satish:
Discussed with @tahini  about the scope of this PR.

As part of this PR, I fixed the issue only for the actual states and excluded GAP states. Since GAP states do not have fixed start and end ranges, the current changes doesn't work on GAP states. I created a new issue #51 for addressing this.

tahini:
Code looks good, but I don't know if it fixes #50. In your screencast, you also select a gap state. Can you reproduce bug #50 with a normal state?

muddana-satish:
> Code looks good, but I don't know if it fixes #50. In your screencast, you also select a gap state. Can you reproduce bug #50 with a normal state?

I'm sorry for the confusion. I added a new screencast showing two issues that this PR fixes.

muddana-satish:
> I could reproduce the bug and it works for normal states.
> 
> As a suggestion when/if you're going to look at the gap states also, instead of row id, start, end to compare with previous selection, maybe you could use row id and the intersection of the currently selected time instead (but I have no idea how the selected time works, so maybe I'm completely wrong)

The problem with this approach is say at initial zoom level we have selected a GAP state. As you zoom in, the same GAP state can get split into two or more GAP states and it's possible that no state has intersection in our selection range. 

I believe splitting the same state into multiple states based on the zoom level would make it difficult to identify the initial selected state. I would have to check this behavior in trace compass.

MatthewKhouzam (approved):

PR #56, opened 2020-10-09
Fix Timegraphchart's shift+mousewheel and ctrl+mousewheel

muddana-satish:
ctrl+mousewheel and shift+mousewheel should work even when there is no foucs


muddana-satish:
Thanks @PatrickTasse for pointing out this issue and suggesting the fix.

PatrickTasse (approved):

PR #57, opened 2020-10-13
Handle onClick event on timegraph row

### LINKED ISSUE #52: Ability to select a row using Mouse.
User should be able to click on a row to select it.

Select Row

muddana-satish:
This feature was woking fine with earlier versions of PIXI
Current PIXI is not able to handle events for elements with opacity 0.
Changed the Opacity as a workaround fix.

Fixes #52 

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -136,7 +136,8 @@ const timeGraphChart = new TimeGraphChart('timeGraphChart', {
     rowStyleProvider: (row: TimelineChart.TimeGraphRowModel) => {
         return {
             backgroundColor: 0xe0ddcf,
-            backgroundOpacity: row.selected ? 0.6 : 0,
+            // PIXI 5 doesn't handle events on elements with opacity 0. As a workaround set it to 0.001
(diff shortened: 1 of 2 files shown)


tahini:
Adding @MatthewKhouzam as reviewer. I _think_ he worked on something similar at some point when moving to pixi 5, no?

muddana-satish:
Changed Opacity to 0.001 which is almost same as 0. Thanks @MatthewKhouzam  for the suggestion.

MatthewKhouzam (approved):
LGTM, it's exactly the hack we had to due because of Pixi

tahini (approved):
Works well with the example app

muddana-satish:
@tahini can we revert these changes since @MatthewKhouzam has fixed this issue in #70 ? Or should I create a new PR to remove these changes?

tahini:
@muddana-satish sure! You can revert the commit and then do a PR for it ;-) I think it's the best way to go, if we really want to test that it works before merging.

PR #58, opened 2020-10-13
Improve State Navigation using Right/Left Arrow Buttons

muddana-satish:
Right Arrow selects next available state.
Left Arrow selects previous available state.
shift+Right/shift+Left selects entire area until next/prev available state.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -87,59 +87,49 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
     }
 
     protected maybeCenterCursor() {
-        const selection = this.unitController.selectionRange;
-        const view = this.unitController.viewRange;
-        if (selection && (selection.start < view.start || selection.start > view.end)) {
-            this.centerCursor();
+        if (this.unitController.selectionRange) {
+            const cursorPosition = this.unitController.selectionRange.end;
+            if (cursorPosition < this.unitController.viewRange.start || cursorPosition > this.unitController.viewRange.end) {
+                this.centerCursor();
+            }
         }
     };
 
     protected navigateOrSelectLeft() {
         const row = this.rowController.selectedRow;
-        if (row) {
-            const states = row.states;
-            const nextIndex = states.findIndex((rowElementModel: TimelineChart.TimeGraphRowElementModel) => {
-                const selStart = this.unitController.selectionRange ? (this.shiftKeyDown ? this.unitController.selectionRange.end : this.unitController.selectionRange.start) : 0;
-                return rowElementModel.range.start >= selStart;
-            });
-            let newPos = 0;
-            let elIndex = 0;
-            if (states.length == 0)
-                return false;
-            if (nextIndex > 0) {
-                elIndex = nextIndex - 1;
-            } else if (nextIndex === -1) {
-                elIndex = states.length - 1;
-            }
-            newPos = states[elIndex].range.start;
-            if (this.unitController.selectionRange && this.shiftKeyDown) {
-                this.unitController.selectionRange = { start: this.unitController.selectionRange.start, end: newPos };
+        if (row && this.unitController.selectionRange) {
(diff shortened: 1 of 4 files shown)


tahini:
> Right+shift and Left+shift works correctly.
> 
Please define "correctly"?

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -179,11 +169,26 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     centerCursor() {
         if (this.unitController.selectionRange) {
-            const cursorPosition = this.unitController.selectionRange.start;
+            const cursorPosition = this.unitController.selectionRange.end;
             const halfViewRangeLength = this.unitController.viewRangeLength / 2;
+            let startViewRange = cursorPosition - halfViewRangeLength;
+            let endViewRange = cursorPosition + halfViewRangeLength;
+
+            let delta = 0;
```
tahini, line 177: You don't need the delta variable here, you can just do the operations directly in each of the blocks

tahini (approved):
Works well! Thanks!

I have a minor comment on the code. Also, when we need performance improvement, a binary search could be used instead of a find, but it's fine for now.

timeline-chart/src/layer/time-graph-chart-cursors.ts
muddana-satish: updated. Thanks.

muddana-satish:
> > Right+shift and Left+shift works correctly.
> 
> Please define "correctly"?

Updated the commit message.

tahini:
Here's some issue I noticed while testing with the example timeline. Look at the following screencast, where I do left/right with one of the row with 2 states. I just click right till the end, then come back left till the end, again and again. Notice that the zoom level seems to change, or at least the state gets bigger when I come back to it later, until at some point they don't appear anymore, I just go from far right to far left and back.

leftRightTimeline

muddana-satish:
@tahini  Thank you for catching this. I broke this while trying to address your suggestion on removing 'delta'. I fixed it. Thanks.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -207,6 +210,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     }
 
     protected async maybeFetchNewData(update?: boolean) {
+        console.log("Jaffa");
```
tahini, line 213: translate? :p

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -269,6 +276,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     }
 
     protected addRow(row: TimelineChart.TimeGraphRowModel, height: number, rowIndex: number) {
+        console.log("Po ra Daffa Nuvvu cheppedhi nenenti vinedhi");
```
tahini, line 279: 2 possible translations: "Po ra daffa I hear what you say" or "Poe's what i have to say you're wrong, vinedi". Sounds right?

tahini (commented):
Works well, but some debug artefacts remain...

timeline-chart/src/layer/time-graph-chart.ts
muddana-satish: Sorry for the mess. Cleared it. :/

tahini (approved):
Thanks!

bhufmann (approved):
Looks good to me! Thanks

timeline-chart/src/layer/time-graph-chart.ts
mad-season: Ha. Satish dropped some Telugu on us. @muddana-satish do at least tell us what it means. The suspense is killing me

muddana-satish: It means: _Go Away! Why should I listen to you._ 
That's me interacting with JS engine.

bhufmann:
Thanks for the contribution!

PR #60, opened 2020-10-15
Fix label overflow from the state

### LINKED ISSUE #59: Text value of the BitMapText object overflows
Steps to reproduce: Expand the state's width to the max so that the entire text is visible and then slowly reduce the width of the state's label by scrolling left or right. The text will start to overflow.

ankusht-work:
Fixes #59  

### DIFF
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -56,15 +56,15 @@ export class TimeGraphRowElement extends TimeGraphComponent {
     renderLabel() {
         const position = this._options.position;
         const width = this._options.width;
+        const textWidth = this.textWidth;
+        const labelText = this._model.label;
+        const fontName = TimeGraphRowElement.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2);
 
         if (this.displayObject.children.length) {
             let textObj = this.displayObject.getChildAt(0) as PIXI.BitmapText;
             textObj.x = position.x;
             textObj.y = position.y;
         }
-
-        const textWidth = this.textWidth;
-        const labelText = this._model.label;
         if (textWidth && labelText) {
             if (width <= 0.1 * textWidth && this.displayObject.children.length) {
                 this.displayObject.removeChildAt(0);
@@ -87,14 +87,12 @@ export class TimeGraphRowElement extends TimeGraphComponent {
                     }
                 }
 
-                if (this.displayObject.children.length === 0) {
-                    const fontName = TimeGraphRowElement.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2);
-                    const text = new PIXI.BitmapText(labelText, { fontName: fontName });
-                    this.displayObject.addChild(text);
+                if (this.displayObject.children.length !== 0) {
+                    this.displayObject.removeChildAt(0);
                 }
 
+                this.displayObject.addChild(new PIXI.BitmapText(textStr, { fontName: fontName }));
                 let textObj = this.displayObject.getChildAt(0) as PIXI.BitmapText;
-                textObj.text = textStr;
                 textObj.x = textObjX;
                 textObj.y = textObjY;
             }


ankusht-work:
I am not able to reproduce this issue with 'tracecompass-test-traces' but I think it makes more sense to clear the label and redraw it just like how we do it for time-graph rows and states.

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -55,16 +55,16 @@ export class TimeGraphRowElement extends TimeGraphComponent {
 
     renderLabel() {
         const position = this._options.position;
-        const width = this._options.width;
+        const width = Math.round(this._options.width);
```
muddana-satish, line 58: Any reason why we have to round the width?

ankusht-work: No specific reason just to make the calculation easier (we compare the text width to the 1/10 th of the state's width) but reading through the code again I realized that by rounding off the width we will wrongfully offset the textObjx position since its a const.

MatthewKhouzam (commented):
Some minor issues: 

1- the commit message should be 

`No longer overflow text value of labels

Fixes issue #59 

Signed-off-by: Ankush...`

that way we don't need an internet connection to know what the commit message is.

2- minor: could you take a look at the time graph view implementation in trace compass? We add an ellipsis (...) that makes it clear there's more text. In my testing, the text gets truncated... This could lead to some unfortunate misunderstandings.

mad-season:
if the ellipsis has gone MIA, then we have a regression.
@ankusht-work sync up with me.

ankusht-work:
> 2- minor: could you take a look at the time graph view implementation in trace compass? We add an ellipsis (...) that makes it clear there's more text. In my testing, the text gets truncated... This could lead to some unfortunate misunderstandings.

Hmm, I see the ellipsis. Also, I haven't really made any changes around the existing ellipses logic so it should work. Could you please provide a screenshot?

bhufmann:
It seems to be a problem with the Canvas renderer and not with the openGL2 renderer. I was able to reproduce it with LTTng UST trace and the Flame chart after forcing to use the canvas renderer on my laptop. It also seems that the label has to be certain length to trigger this issue and if there are no states beside the "overflowing" state. I haven't seen it with the Resources View.

image

Note, that the Flame Chart is not working properly but it's not related to this PR.

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -87,14 +87,12 @@ export class TimeGraphRowElement extends TimeGraphComponent {
...
 
-                if (this.displayObject.children.length === 0) {
-                    const fontName = TimeGraphRowElement.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2);
-                    const text = new PIXI.BitmapText(labelText, { fontName: fontName });
-                    this.displayObject.addChild(text);
+                if (this.displayObject.children.length !== 0) {
+                    this.displayObject.removeChildAt(0);
                 }
 
+                this.displayObject.addChild(new PIXI.BitmapText(textStr, { fontName: fontName }));
```
bhufmann, line 94: With this for every redraw of a label a new BitmapText is created and the previous is removed. I would prefer to just update the text and reuse the BitmapText object. It would be good to investigate why the canvas renderer has trouble with it and we can find a different workaround. If not we still can consider this approach later on.

muddana-satish: So by looking at the overall design, we remove all the drawings on the canvas (like rows, states, arrows) and then redraw it again whenever there is an update (like zoom-in/out, changing width). So in my opinion, it makes more sense to do the same for the labels as well. (i.e clear the label every time we update and redraw again).

ankusht-work: > With this for every redraw of a label a new BitmapText is created and the previous is removed. I would prefer to just update the text and reuse the BitmapText object. It would be good to investigate why the canvas renderer has trouble with it and we can find a different workaround. If not we still can consider this approach later on.

Yes, we are removing the label and re-creating it again. As you said, the update works fine in the case of WebGL but somehow behaves differently for canvas-renderer. I found a somewhat related issue on the pixi repo https://github.com/pixijs/pixi.js/issues/4085.

I haven't been able to find the exact root cause for such a sporadic behavior but are you ok with this change until we come up with something more concrete? @MatthewKhouzam

MatthewKhouzam: At the very least, let's have a cache or something. Thoughts?

MatthewKhouzam (changes requested):
Adding this light -1 until we discuss the point

timeline-chart/src/components/time-graph-row-element.ts
MatthewKhouzam: Ankush and I discussed offline (online) and he will try to make a pixibitmap repo. Maybe the bitmap objects don't need to be re-created, but only replaced for the draw to be correct. Either way, this is another limitation of PIXI.

MatthewKhouzam: We have done some in depth study of the bitmaptext.

The conclusion: bitmaptext caches the font in a static cache. Therefore we can new BitmapText all day long with minimal performance impact! :) :) :) Feel free to merge it!

MatthewKhouzam (approved):

MatthewKhouzam (changes requested):
Reminder, fix the commit message

ankusht-work:
> Reminder, fix the commit message

Fixed the commit message. Please take a look.

MatthewKhouzam (approved):

MatthewKhouzam:
For the other reviewers: the if the font doesn't change in the current implementation of pixi, we won't get a performance boost by re-using objects. The reason: each leter is rendered to a bitmap and then it's reassembled, guttenberg style!

bhufmann:
> For the other reviewers: the if the font doesn't change in the current implementation of pixi, we won't get a performance boost by re-using objects. The reason: each leter is rendered to a bitmap and then it's reassembled, guttenberg style!

Thanks for the investigation and updates. This addresses my earlier concern.

bhufmann (approved):
Looks good to me. Thanks!

PR #67, opened 2020-10-23
Reset fetching flag in a try-finally statement

### LINKED ISSUE #66: Uncaught exception causes prevents further query to server to happen
The flag "this.fetching" can be set to true and won't be reset if an exception created in between (see lines below in time-graph-chart.ts).

BR
Bernd

           `this.fetching = true;
            const rowData = await this.providers.dataProvider(viewRange, resolution);
            if (rowData) {
                this.providedResolution = rowData.resolution;
                this.providedRange = rowData.range;
                this.setRowModel(rowData.rows);
                this.removeChildren();
                this.addRows(this.rows, this.rowController.rowHeight);
            }
            this.fetching = false;`

bhufmann:
Fixes #66

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -188,18 +188,22 @@ export class TimeGraphChart extends TimeGraphChartLayer {
             update
         )) {
             this.fetching = true;
-            const rowData = await this.providers.dataProvider(viewRange, resolution);
-            if (rowData) {
-                this.providedResolution = rowData.resolution;
-                this.providedRange = rowData.range;
-                this.setRowModel(rowData.rows);
-                this.removeChildren();
-                this.addRows(this.rows, this.rowController.rowHeight);
-                if (this.isNavigating) {
-                    this.selectStateInNavigation();
+            try {
+                const rowData = await this.providers.dataProvider(viewRange, resolution);
+                if (rowData) {
+                    this.providedResolution = rowData.resolution;
+                    this.providedRange = rowData.range;
+                    this.setRowModel(rowData.rows);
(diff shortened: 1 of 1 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -219,18 +219,21 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                    this.providedResolution = rowData.resolution;
+                    this.providedRange = rowData.range;
+                    this.setRowModel(rowData.rows);
+                    this.removeChildren();
+                    this.addRows(this.rows, this.rowController.rowHeight);
+                    if (this.isNavigating) {
+                        this.selectStateInNavigation();
+                    }
                 }
+            } catch (error) {
```
bhufmann, line 234: should be finally :-)

PatrickTasse (approved):

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -219,18 +219,21 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                    this.providedRange = rowData.range;
+                    this.setRowModel(rowData.rows);
+                    this.removeChildren();
+                    this.addRows(this.rows, this.rowController.rowHeight);
+                    if (this.isNavigating) {
+                        this.selectStateInNavigation();
+                    }
                 }
+            } finally {
+                this.fetching = false;
```
muddana-satish, line 235: though not related, can we also set this.isNavigating to 'false' to make sure we set all the flags correctly.

bhufmann: make sense

PR #70, opened 2020-10-23
Make all opacity levels work with PIXIjs 5.0

MatthewKhouzam:
Clamp values to range that is acceptable to PIXIjs 5


muddana-satish:
We can remove changes from #57 if we're going with this PR.

tahini (approved):
Very nice, so when PIXI supports selecting 0 opacity states, we can easily change this, and this makes PR https://github.com/theia-ide/theia-trace-extension/pull/150 unnecessary. Me gusta!

MatthewKhouzam:
@tahini the main reason is this is a PIXI issue. so I want to decouple it from our implementation. If we want to step away from that library it'll be easier.

tahini:
@MatthewKhouzam Are you going to merge this? Or are you waiting for something?

MatthewKhouzam:
Oops? :)

PR #72, opened 2020-11-10
Revert: Handle onClick event on timegraph row

muddana-satish:


tahini (approved):
Thanks!

MatthewKhouzam (approved):
SWEET!

PR #73, opened 2020-11-10
add vertical space in-between for states on adjacent rows

maria-ericsson:


maria-ericsson:
**Existing view**:
existing



**After adding vertical space for rows:**
with_vertical space

PR #74, opened 2020-11-11
add vertical space in-between for states on adjacent rows

maria-ericsson:

### DIFF
--- a/timeline-chart/src/components/time-graph-row-element.ts
+++ b/timeline-chart/src/components/time-graph-row-element.ts
@@ -30,6 +30,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
     ) {
         super(id, displayObject);
         this.height = _style.height || 14;
+        this.height = Math.min(this.height, _row.height -1);
         this.position = {
             x: this.range.start,
             y: this._row.position.y + ((this.row.height - this.height) / 2)


timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -16,6 +16,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
     height: number;
     position: TimeGraphElementPosition;
     textWidth: number = 0;
+    MAX_HEIGHT: number = 19;
```
tahini, line 19: Where do 19 and 14 below come from? Just arbitrary value to make it work? If so, maybe you could add a fixme to improve general height calculation. I seem to recall having heard about "rows" that would span like 2 rows or that would have custom sizes eventually? Those hard-coded values will have to be updated then.

tahini (approved):
It gives a nicer look to the thread status view, especially the set of green swappers at the top, they look much nicer :smile:

timeline-chart/src/components/time-graph-row-element.ts
maria-ericsson: I set the maximum state height to "1" less than the row height of the timeline chart. Row height for timeline chart is 20. That was the reason for setting up MAX_HEIGHT to 19. I replaced the hard coded value "19" by "_row.height -1" and pushed the code again.

tahini (approved):
Even better!

MatthewKhouzam (approved):
Looks great!

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -30,6 +30,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
     ) {
         super(id, displayObject);
         this._height = _style.height || 14;
+        this._height = Math.min(this._height, _row.height -1);
```
bhufmann, line 33: Can _row.height be 0? This would result in this._height being negative value. Maybe we should have clamp it to 0.

bhufmann (changes requested):

timeline-chart/src/components/time-graph-row-element.ts
maria-ericsson: Added the check for  _row.height equal to 0

timeline-chart/src/components/time-graph-row-element.ts
```diff
@@ -30,6 +30,7 @@ export class TimeGraphRowElement extends TimeGraphComponent {
     ) {
         super(id, displayObject);
         this._height = _style.height || 14;
+        this._height = _row.height === 0? 0 : Math.min(this._height, _row.height -1);
```
bhufmann, line 33: Small formatting comments. Please add a space before ? and 1 
        this._height = _row.height === 0 ? 0 : Math.min(this._height, _row.height - 1); 

Sorry, I missed that earlier.

bhufmann (changes requested):
Looks good to me

timeline-chart/src/components/time-graph-row-element.ts
maria-ericsson: Thank you! Done.

bhufmann (approved):
Thanks for the contribution!

bhufmann:
@maria-ericsson could you please rebase the patch on latest master so that I can merge it?

bhufmann (changes requested):
I've noticed that you merged master into you branch and then updated the merge commit to fix my latest comments. Please update your branch so that it is rebased on latest master and contains only 1 commit with the changes? This will keep the git history linear. Thanks, Bernd

bhufmann (approved):
Ok, let's do a merge commit instead this time. Thanks for the contribution.

PR #76, opened 2020-11-19
add markers as new layer to timeline chart

### LINKED ISSUE #75: Show annotation 'ticks' in timeline chart
Allow exemplars/regions of interest to be shown in the timeline chart.

Examples could be a system crash/core dump or simply an event that may be a trigger to show a change in state. (See instant events in the event-trace format)

MatthewKhouzam:
Add support marker layer with the following styles:

* 'cross'
* 'diamond'
* 'circle'

Size and color can be defined.

Fixes #75

### DIFF
--- a/example/src/test-data-provider.ts
+++ b/example/src/test-data-provider.ts
@@ -64,6 +64,11 @@ export namespace TestData {
          * Array of states
          */
         states: TimeGraphState[];
+
+        /**
+         * Array of markers
+         */
+        annotations: TimeGraphAnnotation[];
     }
 
     /**
@@ -87,6 +92,23 @@ export namespace TestData {
          */
         value: number;
 
+    }
+
+    /**
+      * Time graph state
+      */
+    export interface TimeGraphAnnotation {
+        /**
+         * Start time of the state
+         */
+        startTime: number;
+
+        duration: number;
+
+        /**
+         * Label to apply to the state
+         */
+        label: string | null;
     }
 
     /**
@@ -163,8 +185,9 @@ export class TestDataProvider {
         const rows: TimelineChart.TimeGraphRowModel[] = [];
         const range = opts.range || { start: 0, end: this.totalLength };
         const resolution = opts.resolution || this.totalLength / this.canvasDisplayWidth;
-        timeGraphEntries.model.entries.forEach((entry: any, rowIndex: number) => {
-            const states: TimelineChart.TimeGraphRowElementModel[] = [];
+        timeGraphEntries.model.entries.forEach((entry: any, rowIndex: number): void => {
+            const states: TimelineChart.TimeGraphState[] = [];
+            const annotations: TimelineChart.TimeGraphAnnotation[] = [];
             const row = timeGraphStates.model.rows.find(row => row.entryID === entry.id);
             let hasStates = false;
             let prevPossibleState = 0;
@@ -191,6 +214,24 @@ export class TestDataProvider {
                         nextPossibleState = state.startTime + state.duration - entry.startTime;
                     }
                 });
+
+                const _annotations = row.annotations;
+                if (!!_annotations) {
+                    _annotations.forEach((annotation: any, annotationIndex: number) => {
+                        const start = annotation.range.start - entry.startTime;
+                        if (range.start < start && range.end > start) {
+                            annotations.push({
+                                id: 'mark_' + rowIndex + '_' + annotationIndex,
+                                range: {
+                                    start: annotation.range.start - this.absoluteStart,
+                                    end: annotation.range.end - this.absoluteStart
+                                },
+                                label: '',
+                            });
+                        }
+                    });
+                }
+
             }
             rows.push({
                 id: entry.id,
@@ -200,6 +241,7 @@ export class TestDataProvider {
                     end: entry.endTime - entry.startTime
                 },
                 states,
+                annotations,
                 data: {
                     type: entry.type,
                     hasStates
--- a/timeline-chart/src/components/time-graph-annotation.ts
+++ b/timeline-chart/src/components/time-graph-annotation.ts
@@ -1,0 +1,143 @@
+import * as PIXI from "pixi.js-legacy"
+import { TimeGraphComponent, TimeGraphElementPosition, TimeGraphComponentOptions } from "./time-graph-component";
+import { TimeGraphRow } from "./time-graph-row"
+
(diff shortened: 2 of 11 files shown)


timeline-chart/src/time-graph-model.ts
```diff
@@ -17,6 +17,7 @@ export namespace TimelineChart {
         name: string
         range: TimeGraphRange
         states: TimeGraphRowElementModel[]
+        annotations: TimeGraphAnnotation[]
```
PatrickTasse, line 20: Should this be { [category: string]: TimeGraphAnnotation[] }?

Do we want to group annotations per category, or do we want to add the category in each annotation, or not have the category at all?

For context, the intended purpose of annotation category is 1) allow bulk show/hide per category 2) to be shown in tooltip 3) not supported yet but to only request and receive shown categories in TSP

timeline-chart/src/time-graph-model.ts
```diff
@@ -38,4 +39,8 @@ export namespace TimelineChart {
         data?: { [key: string]: any }
         // Q: Can I select an arrow?
     }
+
+    export interface TimeGraphAnnotation extends TimeGraphRowElementModel {
```
PatrickTasse, line 43: Maybe this can be its own class (not extending), I don't think we need 'id' or 'selected' from the parent. Unless we support selecting an annotation. We don't need to have 'sourceId', it is the 'id' of the TimeGraphRowModel to which this annotation belongs.

We need 'type' (either CHART or TREE) to know if annotation is to be shown on the chart or in the tree (unless the tree is a widget external to timeline-chart? in which case only the CHART annotations would end up here). Alternatively, perhaps the 'range' can be optional, and if not present it means annotation goes on the tree.

timeline-chart/src/time-graph-model.ts
MatthewKhouzam: Should it not be done on the model side? I want to keep the view dumb.

timeline-chart/src/time-graph-model.ts
MatthewKhouzam: I think it's always Chart as tree is the way to go, since the tree is provided by theia.

timeline-chart/src/time-graph-model.ts
PatrickTasse: Yes, after discussion, timeline-chart doesn't need to know the category, any filtering on category or tooltip should be done by the user of the chart.

timeline-chart/src/time-graph-model.ts
PatrickTasse: OK, the annotations with type 'TREE' will be filtered-out by the user and only those that are 'CHART' will be put in this model, so we don't need type.

I think we don't need sourceId, so perhaps we don't need a new class at all, annotations can fit in the TimeGraphRowElementModel. Unless it can be useful for 'instance of' checks?

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -0,0 +1,90 @@
+import * as PIXI from "pixi.js-legacy"
+
+import { TimeGraphComponent, TimeGraphElementPosition, TimeGraphComponentOptions } from "./time-graph-component";
+
+export interface TimeGraphAnnotationCoordinates extends TimeGraphComponentOptions {
```
PatrickTasse, line 5: We might not need this, if annotations can be a TimeGraphRowElement or a sublass of it.

But then 'symbol' should be added to TimeGraphRowElementStyle.

Should row element be a parent of two classes, state and annotation? Or a single class that does both? What we decide should probably apply to both the *Component and the *Model, and perhaps *Style.

PatrickTasse (changes requested):

example/src/test-data-provider.ts
```diff
@@ -191,6 +214,25 @@ export class TestDataProvider {
...
+                        const start = annotation.range.start - entry.startTime;
+                        if (range.start < start && range.end > start) {
+                            annotations.push({
+                                id: 'mark_' + rowIndex + '_' + annotationIndex,
+                                range: {
+                                    start: annotation.range.start - this.absoluteStart,
+                                    end: annotation.range.end - this.absoluteStart
+                                },
+                                label: '',
+                                data: {type: 'cross, I suppose?',}
```
PatrickTasse, line 230: This is supposed to be {style: {parentKey: 'styleKey', values: {}}}
where 'styleKey' is defined in the style map returned by the /style endpoint, which contains symbol-type, color and height.

MatthewKhouzam: I did something deliberately bad to have a test of the default method... what do you suggest?

PatrickTasse: The style is what identifies it as a symbol. If style or symbol-type property is not present, it's supposed to mean 'none' e.g. not a symbol annotation. If color property is absent it should be #000000. If height property is absent it should be 1.0f e.g. symbol size (radius) = half of row height.

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+        super.update();
+    }
+
+    render(): void {
+        const { symbol } = this._style as TimeGraphAnnotationStyle;
+        const size = this._size;
+        const x = this._options.position.x - (size * 0.5);
+        const y = this._options.position.y - (size * 0.5);
+        if (symbol == 'circle') {
```
PatrickTasse, line 56: In the other libraries that we use, the symbol size is the radius, not the circumference. Would it be better to do the same?

Also, I think it would be more intuitive if the (x, y) given to the draw methods is the center point, not the top-left corner.

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+        } else if (symbol == 'cross') {
+            this.drawCross(x, y, size);
+        } else if (symbol == 'plus') {
+            this.drawPlus(x, y, size);
+        } else if (symbol == 'diamond') {
+            this.drawDiamond(x, y, size);
+        }
+    }
+
+    private drawCircle(x: number, y: number, size: number): void {
```
PatrickTasse, line 67: The circle is drawn in the top-left quadrant of the given point.

PatrickTasse (approved):

MatthewKhouzam:
Thanks for the review!

PR #82, opened 2020-12-15
Fix publishing with lerna v3.22.1

bhufmann:

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "lerna": "3.22.1",
-  "version": "0.1.1",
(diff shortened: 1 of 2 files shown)


lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "lerna": "3.22.1",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
bhufmann, line 3: should be the previous version

MatthewKhouzam (approved):
Go for it! looks about right. (--preid)

PR #85, opened 2020-12-23
Sync event timestamp with timegraph

muddana-satish:

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -86,7 +86,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
         this.unitController.onSelectionRangeChange(() => this.update());
     }
 
-    protected maybeCenterCursor() {
+    public maybeCenterCursor() {
         if (this.unitController.selectionRange) {
             const cursorPosition = this.unitController.selectionRange.end;
             if (cursorPosition < this.unitController.viewRange.start || cursorPosition > this.unitController.viewRange.end) {


bhufmann:
@muddana-satish I currently reviewing this patch. I'll get back to you soon.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -448,13 +453,43 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
     protected selectStateInNavigation() {
         const row = this.rowController.selectedRow;
         if (row && this.unitController.selectionRange) {
             const cursorPosition = this.unitController.selectionRange.end;
-            const rowElement = row.states.find((rowElementModel: TimelineChart.TimeGraphState) => rowElementModel.range.start === cursorPosition || rowElementModel.range.end === cursorPosition);
+            const rowElement = row.states.find((rowElementModel: TimelineChart.TimeGraphState) => rowElementModel.range.start === cursorPosition
+                || rowElementModel.range.end === cursorPosition);
             this.selectRowElement(rowElement);
         }
-        this.setNavigationFlag(false);
```
bhufmann, line 458: This line was accidentally removed

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -86,7 +86,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
         this.unitController.onSelectionRangeChange(() => this.update());
     }
 
-    protected maybeCenterCursor() {
+    public maybeCenterCursor() {
```
bhufmann, line 89: Could please separate this into a separate patch, so that we can get the time sync merged first?
Then we can tackle the row synchronization.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -448,13 +453,43 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
 
+    setEventClickedFlag(flag: boolean) {
+        this.isEventClicked = flag;
+    }
+
+    private selectStateOnEventClicked() {
+        const selectedRow = this.rows.find(row => {
+            if (row && this.unitController.selectionRange) {
+                const cursorPosition = this.unitController.selectionRange.end;
+                const rowElement = row.states.find((rowElementModel: TimelineChart.TimeGraphState) => rowElementModel.range.start === cursorPosition
```
bhufmann, line 464: I think we need to find a better way to sync the row than relying the on the timestamp. With nanosecond precision and a huge number of events it will be click on the trace event of start or end state.
I think the row to select needs to determined outside the library and the library needs to provide an API to show (reveal) the selected row.

bhufmann (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
muddana-satish: It's no longer required after #67. But I decided not to modify this file as part of this PR.

bhufmann (approved):
Looks good to me!

MatthewKhouzam (approved):

PR #87, opened 2020-12-24
Set limit on zoom-in factor

### LINKED ISSUE #86: Browser freezes when zoom-in completely
Kapture 2020-12-24 at 12 38 29

muddana-satish:
fixes #86

### DIFF
--- a/timeline-chart/src/components/time-graph-axis-scale.ts
+++ b/timeline-chart/src/components/time-graph-axis-scale.ts
@@ -138,9 +138,11 @@ export class TimeGraphAxisScale extends TimeGraphComponent {
         if (end > this.unitController.absoluteRange) {
             end = this.unitController.absoluteRange;
         }
-        this.unitController.viewRange = {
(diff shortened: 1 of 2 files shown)


muddana-satish:
Doesn't allow the zoom once it reaches the limit. (In our case if the start and end are the same).

Kapture 2020-12-24 at 19 50 11

mirsky:
Thanks for the patch, works well for mouse scroll and keyboard shortcuts.
Looks like it doesn't cover zooming-in by dragging the time axis though.

muddana-satish:
> Thanks for the patch, works well for mouse scroll and keyboard shortcuts.
> Looks like it doesn't cover zooming-in by dragging the time axis though.

Thank you for pointing it out. Fixed it.

bhufmann (approved):
Thanks for the fix. No more frozen browser.

MatthewKhouzam (approved):

PR #91, opened 2021-01-05
Draw plus annotation symbol for unknown symbol styles

### LINKED ISSUE #90: Provide default annotation symbol
Right now when the annotation symbol style is not known then the annotation symbol is not drawn. There should be a default symbol drawn in for unknown styles (e.g. plus).

bhufmann:
Fixes #90

### DIFF
--- a/timeline-chart/src/components/time-graph-annotation.ts
+++ b/timeline-chart/src/components/time-graph-annotation.ts
@@ -53,6 +53,10 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
         const size = this._size;
         const x = this._options.position.x;
         const y = this._options.position.y;
+
+        if (symbol === undefined || symbol == 'none') {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -61,6 +61,8 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
             this.drawPlus(x, y, size);
         } else if (symbol == 'diamond') {
             this.drawDiamond(x, y, size);
+        } else {
```
PatrickTasse, line 64: We have to make sure that when the symbol value is 'none' or undefined it draws nothing.

Then I'm not sure if it's better to have a default drawing only for known but unsupported symbols, or for any other symbol value.

PatrickTasse (changes requested):

PatrickTasse (approved):

PR #100, opened 2021-01-18
Resize background as well when resizing canvas.

### LINKED ISSUE #99: Background of chart not resized when resizing the window
When resizing the TimeGraphContainer using method reInitCanvasSize(newWidth, newHeight), all the layers are updated but not the background layer. This leads to incorrect visualizations of the data in the new added screen area. 

See https://github.com/theia-ide/theia-trace-extension/issues/209#issue-782150707 for more details on it.

bhufmann:
Fixes #99

### DIFF
--- a/timeline-chart/src/time-graph-container.ts
+++ b/timeline-chart/src/time-graph-container.ts
@@ -25,6 +25,8 @@ export class TimeGraphContainer {
 
     private application: PIXI.Application;
 
+    private background: TimeGraphRectangle;
+
     constructor(protected config: TimeGraphContainerOptions, protected unitController: TimeGraphUnitController, protected extCanvas?: HTMLCanvasElement) {
         let canvas: HTMLCanvasElement
         if (!extCanvas) {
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Makes great sense! Also works! Thanks Bernd!

timeline-chart/src/time-graph-container.ts
```diff
@@ -81,10 +83,15 @@ export class TimeGraphContainer {
...
         if (newHeight === undefined) {
             newHeight = this.config.height;
+        } else {
+            this.config.height = newHeight;
         }
+
         this.application.renderer.resize(newWidth, newHeight);
         this.stateController.updateDisplayWidth();
         this.stateController.updateDisplayHeight();
+        this.background.displayObject.height = newHeight;
```
bhufmann, line 93: the width has to be updated as well.

bhufmann: I'll update that patch for that.

PR #101, opened 2021-01-21
Don't scroll chart when all rows are visible

### LINKED ISSUE #53: Mouse wheel makes rows bottom-aligned
When the timeline chart contains less rows than the maximum number of visible rows, and therefore there is no visible vertical scroll bar, then using the mouse wheel to scroll up or down moves the rows from the top to the bottom of the chart.
bottom

bhufmann:
Fixes #53

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -79,6 +79,9 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         }
 
         const moveVertically = (magnitude: number) => {
+            if (this.rowController.totalHeight <= this.stateController.canvasDisplayHeight) {
+                return;
+            }
             let verticalOffset = Math.max(0, this.rowController.verticalOffset + magnitude);
             if (this.rowController.totalHeight - verticalOffset <= this.stateController.canvasDisplayHeight) {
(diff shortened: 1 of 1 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -79,6 +79,9 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         }
 
         const moveVertically = (magnitude: number) => {
+            if (this.rowController.totalHeight <= this.stateController.canvasDisplayHeight) {
```
PatrickTasse, line 82: Somehow this doesn't prevent scrolling if the total height of the tree has been reduced by collapsing some tree elements. Should the totalHeight be recalculated?

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
PatrickTasse: Looks like the totalHeight is set by the user of the timeline-chart, so it would need to update it when a node is toggled.

PatrickTasse (approved):

timeline-chart/src/layer/time-graph-chart.ts
bhufmann: Thanks for investigating this.

MatthewKhouzam (approved):

PR #102, opened 2021-01-26
Make drawing of TimeGraphAxisScale labels optional and fix their y-pos

PatrickTasse:
The y-pos of the text label should be just below the tick line height.

Drawing the labels should be optional and not done for a TimeGraphGrid
which only needs to draw tick (grid) lines at full height.

Fixes https://github.com/theia-ide/theia-trace-extension/issues/243


MatthewKhouzam:
Can you say fixes 243 in theia trace extention, or link the issue?

MatthewKhouzam (approved):

MatthewKhouzam:
Works here!

bhufmann (approved):
Fixes the reported bug!

PR #105, opened 2021-01-27
Fix overlapping TimeGraphAxisScale labels

PatrickTasse:
Calculate the expected label text width to set the minimum step width.

Fixes https://github.com/theia-ide/timeline-chart/issues/103


bhufmann (approved):
Nice fix. Thanks!

PR #106, opened 2021-01-27
Make TimeGraphRow background and gridline spawn full chart width

PatrickTasse:


bhufmann (changes requested):
It works until resizing the graph. Then it's not full chart as before the patch

bhufmann (approved):
Looks good to me. Tested in my local development environment

PR #113, opened 2021-03-01
Use Vertical alignment for annotations

MatthewKhouzam:
Helps for triangles and inverted triangles

### DIFF
--- a/timeline-chart/src/components/time-graph-annotation.ts
+++ b/timeline-chart/src/components/time-graph-annotation.ts
@@ -10,6 +10,7 @@ export interface TimeGraphAnnotationStyle extends TimeGraphComponentOptions {
     symbol?: string
     size?: number
     color?: number
+    verticalAlign?: string
 }
 
 /*
@@ -18,15 +19,15 @@ export interface TimeGraphAnnotationStyle extends TimeGraphComponentOptions {
 export class TimeGraphAnnotationComponent extends TimeGraphComponent {
 
     // TODO: make a map of the display objects
-    // e.g. cross-14-black etc... 
-    // and then use this._displayObject.addChild(cross-14-black) in the draw.
+    // e.g. cross-14-black-middle etc...
+    // and then use this._displayObject.addChild(cross-14-black-middle) in the draw.
     // if performance is an issue
 
     protected _size: number;
 
     constructor(id: string,
         protected _options: TimeGraphAnnotationComponentOptions,
-        protected _style: TimeGraphAnnotationStyle = { color: 0, size: 7, symbol: 'cross' },
+        protected _style: TimeGraphAnnotationStyle = { color: 0, size: 7, symbol: 'cross', verticalAlign: 'middle' },
         protected _row: TimeGraphRow,
         displayObject?: PIXI.Graphics) {
         super(id, displayObject);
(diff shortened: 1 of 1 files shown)


timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -38,16 +39,24 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
     update(opts: TimeGraphAnnotationComponentOptions, style?: TimeGraphAnnotationStyle): void {
         if (opts) {
             this._options.position.x = opts.position.x;
-            this._options.position.y = this._row.position.y + (this._row.height * 0.5);
+            const align = this._style.verticalAlign;
+            this.updateYPosition(align);
         }
         if (style) {
```
PatrickTasse, line 45: I can't find anywhere this method is called with a style parameter? Perhaps this can be removed?

PatrickTasse (changes requested):

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -26,7 +27,7 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
 
     constructor(id: string,
         protected _options: TimeGraphAnnotationComponentOptions,
-        protected _style: TimeGraphAnnotationStyle = { color: 0, size: 7, symbol: 'cross' },
+        protected _style: TimeGraphAnnotationStyle = { color: 0, size: 7, symbol: 'cross', verticalAlign: 'top' },
```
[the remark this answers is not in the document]
MatthewKhouzam: Done!

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -38,16 +39,24 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
...
+            this._style.verticalAlign = style.verticalAlign;
+            const align = this._style.verticalAlign;
+            this.updateYPosition(align);
         }
         super.update();
     }
 
+    private updateYPosition(align: string | undefined) {
+        const offset = align == 'top' ? 0.25 : align == 'bottom' ? 0.75 : 0.5;
+        this._options.position.y = this._row.position.y + (this._row.height * offset);
```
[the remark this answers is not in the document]
MatthewKhouzam: I am putting it to snap to top or bottom of the row now, not the state, as the annotation is a different notion. You OK with that?

timeline-chart/src/components/time-graph-annotation.ts
PatrickTasse: No I mean we always call update(opts) but never update(opts, style). So the if (style) is never used, I think.

timeline-chart/src/components/time-graph-annotation.ts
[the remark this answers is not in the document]
PatrickTasse: Yes, that also is how it's done also in TC.

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -38,16 +39,26 @@ export class TimeGraphAnnotationComponent extends TimeGraphComponent {
...
+            this.updateYPosition();
         }
         super.update();
     }
 
+    private updateYPosition() {
+        const align = this._style.verticalAlign;
+        const size = this._style.size;
+        if (!!size) {
+            const offset = align == 'top' ? size / 2 : align == 'bottom' ? this._row.height - size / 2 : this._row.height;
```
PatrickTasse, line 57: The symbol size is already half the height of the symbol (e.g. radius). So for 'top' it should be size, for 'bottom' it should be height - size, and for 'middle' it should be height / 2?

PatrickTasse (changes requested):

timeline-chart/src/components/time-graph-annotation.ts
MatthewKhouzam: Let's try with numbers:

size = 7, the row and the symbol are size 14.

top = 7
Bottom = 14-7 = 7
center = 14/2 = 7

I went for size/2 so we can be clamped to the limit

But I am not sure.

PatrickTasse: In your example if row and symbol are 14 (meaning full-height symbol) then its size is 7 and the center y should also be 7 for all cases. But the code gives: 3.5, 11.5 and 14.

For half-height symbol with row height 20, the size is 5, we would want: 5, 20-5=15, 20/2=10.

MatthewKhouzam: I am not really convinced here. look at this:

top:
image
center:
image
bottom:
image


vs

top:
image
center:
image
bottom:
image

timeline-chart/src/components/time-graph-annotation.ts
MatthewKhouzam: Fixed, I think.

PatrickTasse (approved):

muddana-satish:
LGTM. Thanks.

PR #114, opened 2021-03-03
Fix horizontal zoom step ratio

PatrickTasse:
The step for zooming by mouse wheel or keyboard is now 80%/125% of
current window range. It no longer depends on mouse cursor y position.

Browser zoom is disabled when using mouse wheel on time axis.


MatthewKhouzam (approved):
I am OK with it, but I want Ankush to agree too.

ankusht-work (approved):
LGTM

PR #115, opened 2021-03-05
Destroy arrows when updating them

arfio:
Arrows are destroyed when the map is updated to not show them and cleanup gpu memory.

Signed-off-by: Arnaud Fiorini <fiorini.arnaud@gmail.com>

### DIFF
--- a/timeline-chart/src/layer/time-graph-layer.ts
+++ b/timeline-chart/src/layer/time-graph-layer.ts
@@ -47,14 +47,14 @@ export abstract class TimeGraphLayer {
     }
 
     protected removeChildren() {
-        this.children.forEach(child => this.layer.removeChild(child.displayObject));
+        this.children.forEach(child => child.destroy());
         this.children = [];
     }
 
     protected removeChild(child: TimeGraphComponent) {
+        child.destroy();
         const idx = this.children.findIndex(c => c === child);
         idx && this.children.splice(idx, 1);
-        this.layer.removeChild(child.displayObject);
     }
 
     protected getPixels(ticks: number) {
--- a/timeline-chart/src/components/time-graph-arrow.ts
+++ b/timeline-chart/src/components/time-graph-arrow.ts
@@ -17,6 +17,11 @@ export class TimeGraphArrowComponent extends TimeGraphComponent {
         this.head = new PIXI.Graphics();
     }
 
+    destroy() {
+        this.head.destroy();
+        super.destroy();
+    }
+
     render(): void {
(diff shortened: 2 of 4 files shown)


MatthewKhouzam (approved):

MatthewKhouzam (approved):

timeline-chart/src/layer/time-graph-layer.ts
```diff
@@ -48,10 +48,12 @@ export abstract class TimeGraphLayer {
 
     protected removeChildren() {
         this.children.forEach(child => this.layer.removeChild(child.displayObject));
+        this.children.forEach(child => child.destroy());
         this.children = [];
     }
 
     protected removeChild(child: TimeGraphComponent) {
+        child.destroy();
```
bhufmann, line 56: this will destroy the child's displayObject. However, it is used later on in line 59. I wonder if it should be removed first, before destroyed? 

I debugged the removeChild(child.displayObject)  with your implementation and after moving the destroy after the removeChild() line. I noticed that inside the method it takes a different code path when stepping through.

bhufmann (changes requested):

timeline-chart/src/layer/time-graph-layer.ts
arfio: I checked the documentation and it seems that calling destroy() on a DisplayObject will automatically remove it from its parent Container. (https://pixijs.download/release/docs/PIXI.DisplayObject.html#destroy). So maybe I should update the patch and remove the this.layer.removeChild() which is no longer needed ?

bhufmann: > I checked the documentation and it seems that calling destroy() on a DisplayObject will automatically remove it from its parent Container. (https://pixijs.download/release/docs/PIXI.DisplayObject.html#destroy). So maybe I should update the patch and remove the this.layer.removeChild() which is no longer needed ?

Yes, this will work too. Thanks for checking.

MatthewKhouzam (commented):
test to say not approved?

MatthewKhouzam:
I need to re-review. Not sure how to unaprove.

bhufmann (approved):
Looks good to me. Thanks.

bhufmann:
@MatthewKhouzam could you please re-review this one?

MatthewKhouzam (approved):
I am OK with this. Works.

MatthewKhouzam (approved):
I am OK with this. Works. (Wish GH would let me review a PR and not commit by commit. I feel like a dweeb)

PR #117, opened 2021-03-26
Add API to get visible entries

MatthewKhouzam:
Allow only fetching data that is shown at the moment.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -43,6 +43,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     protected selectedElementChangedHandler: ((el: TimelineChart.TimeGraphState) => void)[] = [];
     protected providedRange: TimelineChart.TimeGraphRange;
     protected providedResolution: number;
+    protected visibleEntries: number[] = [];
 
     protected fetching: boolean;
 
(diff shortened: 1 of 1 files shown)


MatthewKhouzam:
Not yet ready, showing progress

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -214,8 +216,23 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+
     protected updateScaleAndPosition() {
         if (this.rows) {
+            const canvas = this.canvas;
+            if (!canvas)
+                return;
+            const top = canvas.offsetTop;
+            const height = canvas.offsetHeight;
+            const bottom = top + height;
+            console.log("Bounds " + top + " - " + bottom);
```
PatrickTasse, line 234: remove

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -228,6 +245,12 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                         width: this.stateController.canvasDisplayWidth
                     }
                     comp.update(opts);
+                    const cTop = comp.position.y;
+                    const cBottom = comp.height + cTop;
+                    console.log("Object " + cTop + " - " + cBottom + " / " + top + " - " + bottom);
```
PatrickTasse, line 250: remove

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -228,6 +245,12 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                    if (cTop <= bottom && top <= cBottom) {
```
PatrickTasse, line 251: It appears that the vertical scrolling is not taken into account. The canvas bounds are unaffected by vertical scrolling, and the row y-position is fixed regardless of scrolling (e.g. it is the position in the full-height virtual scrollable area). Therefore it is always the first few rows of the model being considered visible, even when they are scrolled out of visible range.

PatrickTasse (changes requested):

PR #119, opened 2021-04-08
Support horizontal panning with middle button or Ctrl+left button

PatrickTasse:
Allow panning with the mouse with middle button drag or Ctrl+left button
drag. Change the cursor to 'grabbing' while panning is ongoing.

Change the cursor to 'grabbing' when only Ctrl is pressed to indicate
that left button can be used to pan.

Restore the cursor to 'default' when panning is ended by releasing the
mouse button that initiated it, or when the conditions to initiate
Ctrl+left button panning are no longer met.

Prevent time selection with any other button than the left button or
when the Ctrl or Alt key is pressed.

### DIFF
--- a/doc/documentation.md
+++ b/doc/documentation.md
@@ -36,8 +36,10 @@ A reusable time axis component, that can be used independently of the other comp
 * The library provides with a pull hook, to lazily fetch data from any source.
 * Selection of elements (single units) is accessible with a full read/write/listen API.
 * Keystroke handling for navigation and time selection is supported. For a selected row, it's possible to use left/right arrow to go from one element's start time to end and vice versa, as well as skipping to the next/previous element on that row. Using shift + left mouse click will select a time range and the cursors T1 and T2 are drawn accordingly.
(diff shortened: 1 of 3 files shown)


bhufmann (changes requested):
Could you please add a description of panning to https://github.com/theia-ide/timeline-chart/blob/master/doc/documentation.md as well?

bhufmann (changes requested):
I noticed, using CTRL + left mouse, after mouse up, it's not possible to pan and I have to release the CTRL key and press it again to make it work again. I would expect, while having the CTRL key pressed I can pan multiple times. What do you think?

bhufmann (approved):
Looks good to me. Thanks.

MatthewKhouzam (approved):
Works

MatthewKhouzam (approved):

MatthewKhouzam (approved):

MatthewKhouzam (approved):

PR #121, opened 2021-04-14
Add APIs to clean-up components and to remove event handlers

### LINKED ISSUE #98: TimeGraphUnitController handlers cannot be removed
Handlers that are added to the unit controller with onViewRangeChange() and onSelectionRangeChanged() can never be removed.

So if the component that added a handler is unmounted, its handler will continue to be called.

### LINKED ISSUE #120: TimeGraphChart mouse and keyhandlers cannot be removed
Handlers that are added to the canvas in TimeGraphChart with onCanvasEvent(). 

So if the component that added a handler is unmounted, its handler will continue to be called.

bhufmann:
When being destroyed, make sure that all children are removed as well as
all event handlers are removed to avoid

Fixes #98
Fixes #120

### DIFF
--- a/timeline-chart/src/layer/time-graph-axis-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-axis-cursors.ts
@@ -1,10 +1,12 @@
 import { TimeGraphAxisCursor } from "../components/time-graph-axis-cursor";
+import { TimelineChart } from "../time-graph-model";
 import { TimeGraphLayer } from "./time-graph-layer";
 
 export class TimeGraphAxisCursors extends TimeGraphLayer {
     protected firstCursor?: TimeGraphAxisCursor;
     protected secondCursor?: TimeGraphAxisCursor;
     protected color: number = 0x0000ff;
+    private _updateHandler: { (): void; (viewRange: TimelineChart.TimeGraphRange): void; (selectionRange: TimelineChart.TimeGraphRange): void; (viewRange: TimelineChart.TimeGraphRange): void; (selectionRange: TimelineChart.TimeGraphRange): void; };
 
     constructor(id: string, style?: { color?: number }) {
         super(id);
@@ -15,8 +17,9 @@ export class TimeGraphAxisCursors extends TimeGraphLayer {
     }
 
     afterAddToContainer() {
-        this.unitController.onViewRangeChanged(() => this.update());
-        this.unitController.onSelectionRangeChange(() => this.update());
+        this._updateHandler = (): void => this.update();
+        this.unitController.onViewRangeChanged(this._updateHandler);
+        this.unitController.onSelectionRangeChange(this._updateHandler);
         this.update();
     }
 
@@ -59,5 +62,13 @@ export class TimeGraphAxisCursors extends TimeGraphLayer {
             delete this.firstCursor;
             delete this.secondCursor;
         }
+    }
+
+    destroy(): void {
+        if (this.unitController) {
+            this.unitController.removeViewRangeChangedHandler(this._updateHandler);
+            this.unitController.removeSelectionRangeChangedHandler(this._updateHandler);
+        }
+        super.destroy();
     }
 }
--- a/timeline-chart/src/layer/time-graph-chart-arrows.ts
+++ b/timeline-chart/src/layer/time-graph-chart-arrows.ts
@@ -6,9 +6,11 @@ import { TimeGraphChartLayer } from "./time-graph-chart-layer";
 export class TimeGraphChartArrows extends TimeGraphChartLayer {
 
     protected arrows: Map<TimelineChart.TimeGraphArrow, TimeGraphArrowComponent>;
+    private _updateHandler: { (): void; (viewRange: TimelineChart.TimeGraphRange): void; (viewRange: TimelineChart.TimeGraphRange): void; };
 
     protected afterAddToContainer() {
-        this.unitController.onViewRangeChanged(() => this.update());
+        this._updateHandler = (): void => this.update();
+        this.unitController.onViewRangeChanged(this._updateHandler);
 
         this.rowController.onVerticalOffsetChangedHandler(verticalOffset => {
             this.layer.position.y = -verticalOffset;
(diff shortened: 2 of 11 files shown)


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -214,7 +221,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         this.stage.on('mouseup', mouseUpHandler);
```
PatrickTasse, line 221: The mousemove handler added to the stage does not get removed and still gets called after closing the time graph. I would guess all the other stage handlers aren't removed either, however they aren't being triggered when their time graph is closed.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -273,21 +275,28 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+        this.onCanvasEvent('mousewheel', this._mouseWheelHandler);
+        this.onCanvasEvent('wheel', this._mouseWheelHandler);
+        this.onCanvasEvent('contextmenu', this._contextMenuHandler);
 
         this.rowController.onVerticalOffsetChangedHandler(verticalOffset => {
             this.layer.position.y = -verticalOffset;
         });
 
-        this.unitController.onViewRangeChanged(() => {
+        this._viewRangeChangedHandler = () => {
```
PatrickTasse, line 290: This handler doesn't get removed and still gets called after time graph is closed.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -325,6 +334,32 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         }
     }
 
+    destroy() {
```
PatrickTasse, line 337: This method does not get called when the time graph is closed, or when the trace tab is closed, or when the Trace Viewer view is closed.

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
PatrickTasse: Sorry I was missing https://github.com/theia-ide/theia-trace-extension/pull/342. Now it gets called.

timeline-chart/src/layer/time-graph-chart.ts
PatrickTasse: Now it gets removed with https://github.com/theia-ide/theia-trace-extension/pull/342.

timeline-chart/src/layer/time-graph-chart.ts
PatrickTasse: None of the stage listeners get removed. Or is it that the stage itself needs to be destroyed (along with all its listeners)?

timeline-chart/src/layer/time-graph-axis-cursors.ts
```diff
@@ -15,8 +16,8 @@ export class TimeGraphAxisCursors extends TimeGraphLayer {
     }
 
     afterAddToContainer() {
```
PatrickTasse, line 18: This comment applies to:
- TimeGraphAxis
- TimeGraphAxisCursors
- TimeGraphChartGrid
- TimeGraphChartSelectionRange
- TimeGraphNavigator

In these classes, afterAddToContainer() is called when the layer is added and initialized in ReactTimeGraphContainer's componentDidMount(). This adds the listeners for the current instance.

But every time render() is done (by any change of selection range of view range with the mouse), a new instance of these classes is created, many times as a new instance for every pixel mouse move during time selection. But for these instances, afterAddToContainer() is not called and no listeners are added.

So the listener handling happens only for the original instance. This instance gets destroyed when the time graph is closed and the listeners are removed.

Not sure there is an issue here, but I'm wondering what happens with all those instances created by render(). These instances never have their update() method called.

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-axis-cursors.ts
bhufmann: I figured out why the constructor is called at each render. It's caused by how it is used in the https://github.com/theia-ide/theia-trace-extension where they are embedded in a react component and at each render a new instance is created. So, this needs to be fixed there.

timeline-chart/src/layer/time-graph-chart.ts
bhufmann: I'll remove all the stage listeners at destroy, as well, plus destroy the pixi application when closing.

timeline-chart/src/layer/time-graph-chart-arrows.ts
```diff
@@ -64,4 +66,10 @@ export class TimeGraphChartArrows extends TimeGraphChartLayer {
         }
     }
 
+    destroy() : void {
+        super.destroy();
```
PatrickTasse, line 70: In addArrows() above, when we replace all the arrows with a new list, each old arrow is destroyed. But they are not removed as children, so when super.destroy() is called here, they will be destroyed a second time, which throws an exception.

The old arrows should be removed with removeChildren(), this will remove and destroy all of them in one call.

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart-arrows.ts
bhufmann: sounds good. I'll fix it

PatrickTasse (approved):

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -287,4 +301,28 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
             delete this.secondCursor;
         }
     }
+
+    destroy() : void {
+        super.destroy();
```
MatthewKhouzam, line 306: asking, should this be at the end of the "destroy"?

bhufmann: @MatthewKhouzam hmm, interesting point. I think it's better to clean-up this class first before call destroy on parent. I'll change it here and other relevant place in this PR.

MatthewKhouzam (approved):
I'm good with this

PR #122, opened 2021-04-16
Support horizontal zooming selection with right mouse button

PatrickTasse:
Allow zooming selection with the mouse with right button drag. Change
the cursor to 'col-resize' while zooming selection is ongoing.

Restore the cursor to 'default' when zooming selection is ended by
releasing the right mouse button.

Add a zooming selection transparent overlay while zoom selection is
ongoing. Update the overlay when the view range is changed externally.

Prevent the context menu popup on right mouse click on time graph.

Ensure right mouse button mouse up outside of time graph canvas is
properly detected and handled.


bhufmann (changes requested):
I noticed that one has to click with the left mouse button (middle mouse button) first to make the right mouse zooming work.

bhufmann (approved):
Looks good to me. Thanks for the improvement.

MatthewKhouzam (approved):
I cannot approve this hard enough. Great work!

PR #123, opened 2021-04-27
Zooming selection

arfio:
A part of https://github.com/theia-ide/theia-trace-extension/issues/330 
Adds the capability of zooming using the right click & drag.
For now, the zooming action deletes the selection range. It could also work by having an additional selection zone like in Trace Compass. I'm open for extra comments.
output

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -14,6 +14,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
     protected firstCursor?: TimeGraphCursor;
     protected secondCursor?: TimeGraphCursor;
     protected color: number = 0x0000ff;
+    protected zoomSelecting: boolean = false;
 
     constructor(id: string, protected chartLayer: TimeGraphChart, protected rowController: TimeGraphRowController, style?: { color?: number }) {
         super(id, rowController);
@@ -85,7 +86,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
         });
         this.stage.on('mousemove', (event: PIXI.InteractionEvent) => {
             this.mouseButtons = event.data.buttons;
-            if (this.mouseSelecting && this.unitController.selectionRange) {
+            if ((this.zoomSelecting || this.mouseSelecting) && this.unitController.selectionRange) {
                 if ((this.mouseButtons & 1) === 0) {
                     // handle missed button mouseup event
                     this.mouseSelecting = false;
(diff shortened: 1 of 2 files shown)


bhufmann:
@arfio #122 has been merged to provide right-mouse button zooming (CC @PatrickTasse). Could you please try the latest in master and see if it works as expected?

arfio:
I tried with master, I can zoom but I have a few issues (that I may have on this patch also):
- when playing with it, after some time the selection zone disappear and I do not have any visual cues on what range I am selecting

- when selecting over the end, either to the left or the right, the zoom does not work correctly. Instead of taking the limit of the view range and the time selected as the new view range, it zooms towards the center.

Also this patch fixes a potential undefined error for the selectionRange.

PatrickTasse:
> I tried with master, I can zoom but I have a few issues (that I may have on this patch also):
> 
> * when playing with it, after some time the selection zone disappear and I do not have any visual cues on what range I am selecting
> * when selecting over the end, either to the left or the right, the zoom does not work correctly. Instead of taking the limit of the view range and the time selected as the new view range, it zooms towards the center.
> 
> Also this patch fixes a potential undefined error for the selectionRange.

I was able to reproduce this behaviour, I hope it is the same cause that you experienced:

While zoom-selecting with the right mouse button, you can still Ctrl+mouse wheel zoom. If this triggers an update of the row model in TimeGraphChart.maybeFetchNewData() (or maybe some other condition causes this method to be called), all children are removed (the intention is to remove all states, annotations and arrows before fetching them again). This also removes the zoomingSelection rectangle. So updateZoomingSelection() is called to re-create it. That method will first remove the old rectangle. But since it's already been removed by removing all children, then an exception is thrown when calling destroy() on the PIXI object, it is not allowed to call destroy() twice on the same object. Because of the exception, the internal state is wrong and this causes the subsequent zooming to use the wrong time range.

The fix is to delete the zoomingSelection variable in TimeGraphChart.maybeFetchNewData(), just before calling updateZoomingSelection().

PatrickTasse:
The patch doesn't work for me, perhaps because it has been rebased and conflicts with the other implementation now? While holding the right mouse button, mouse move has no effect so when releasing the button it has no effect.

The right button changes the current time selection. When releasing the button, there is no longer any selection in the time chart, but the arrow cursor on the axis remains.

The zooming by selection functionality should not modify the time selection.

bhufmann:
@arfio is this PR still relevant? Otherwise, you can close.

arfio:
I will close it, for now the code in it is not relevant. If I rework this PR in the future, maybe it will be, but it is not a priority

PR #124, opened 2021-04-28
Add support for range events in the timeline chart

ankusht-work:

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -7,13 +7,13 @@ import { TimeGraphContainer } from "timeline-chart/lib/time-graph-container";
 import { TimeGraphChartCursors } from "timeline-chart/lib/layer/time-graph-chart-cursors";
 import { TimeGraphChartSelectionRange } from "timeline-chart/lib/layer/time-graph-chart-selection-range";
 import { TimeGraphAxisCursors } from "timeline-chart/lib/layer/time-graph-axis-cursors";
-// import { timeGraph } from "timeline-chart/lib/test-data";
 import { TimelineChart } from "timeline-chart/lib/time-graph-model";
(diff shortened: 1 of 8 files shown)


MatthewKhouzam (changes requested):
1- the patch is in for opacity
2- please add some examples in the index.ts like the arrows and symbols in the timeline chart.

ankusht-work:
> Done

MatthewKhouzam (commented):
image
I think the annotation is drawn several times... check out the opacity. Also, look at how the top and bottom half of the highlighted row are different.

MatthewKhouzam:
https://user-images.githubusercontent.com/419535/118017032-1bb56780-b324-11eb-86b6-145ebf3affb7.mp4
The markers fail at zooming. :(

ankusht-work:
> image
> I think the annotation is drawn several times... check out the opacity. Also, look at how the top and bottom half of the highlighted row are different.

Thank you for pointing this out. By adding lost event as a new layer to the timeline-chart, it does seem to fix all the issues.

ankusht-work:
> https://user-images.githubusercontent.com/419535/118017032-1bb56780-b324-11eb-86b6-145ebf3affb7.mp4
> The markers fail at zooming. :(


lost_event

MatthewKhouzam:
Double drawn when inited.
vokoscreen-2021-05-21_13-55-54

ankusht-work:
> Double drawn when inited.
> vokoscreen-2021-05-21_13-55-54

Ok, I see what's going on here. In the time-graph-layer class, the addChild method directly calls the "render" instead of calling the update function and hence the double draw. The update method first clears the display object and then renders it.

bhufmann (approved):
Looks good to me!

PR #125, opened 2021-04-29
Fix TypeError exception in TimeGraphChart

PatrickTasse:
When TimeGraphChart.maybeFetchNewData() was called while right-mouse
zooming was in progress (for example by navigating with WASD or
Ctrl+mouseWheel), the zoomingSelection rectangle was removed, then
updateZoomingSelection() would try to remove it again before re-creating
a new one, leading to an exception. Clear the zoomingSelection variable
before calling the method since the call to removeChildren() has already
removed it.


MatthewKhouzam (approved):
Works

bhufmann (approved):
looks good to me

PR #127, opened 2021-05-04
Rename TimeGraphRowElement to TimeGraphStateComponent

PatrickTasse:
Also rename related classes, methods and variables.


MatthewKhouzam (approved):
LGTM

MatthewKhouzam:
Question, when we merge XY charts into this... will it still apply?

bhufmann (approved):
Looks good to me.

PatrickTasse:
> Question, when we merge XY charts into this... will it still apply?

Those line graphs in a row would not be states, they would be something new, perhaps with the same parent class as the states.

MatthewKhouzam:
Approved!

PR #133, opened 2021-05-18
Changes to support tooltip for annotations

PatrickTasse:
Add model to TimeGraphComponent using generics. Store the model for
state, annotation, arrow and row components.

Change TimeGraphStateMouseInteractions to TimeGraphMouseInteractions so
that they can be added to any TimeGraphComponent.

Add element interactions to each annotation component.

Add category to the TimeGraphAnnotation model.

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -97,7 +97,11 @@ const providers = {
     },
     rowAnnotationStyleProvider: (annotation: TimelineChart.TimeGraphAnnotation) => {
         return {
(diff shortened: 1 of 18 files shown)


bhufmann (changes requested):
@PatrickTasse  It would be good to test the mouse handler in the example. Maybe have a text output beside the "reset cursor button". Otherwise it looks good to me.

bhufmann (approved):
LGTM ... thanks a lot

MatthewKhouzam (approved):
Works, I don't like your color choice though

PR #137, opened 2021-06-03
Add range events layer

ankusht-work:


bhufmann (approved):
Looks good to me. Performance issue in previous patch doesn't exist anymore. Thanks!

MatthewKhouzam (approved):
Works here!

PR #140, opened 2021-06-14
Support zooming of timegraph chart from external libs

muddana-satish:


MatthewKhouzam (approved):
Cleaner and does the same thing. Works on Linux.

bhufmann (approved):
Looks good to me

PR #141, opened 2021-06-14
Chart freezes when zooming using right mouse button

muddana-satish:

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -272,10 +272,17 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                 this.mouseEndX = e.offsetX;
                 const start = this.mouseZoomingStart;
                 const end = this.unitController.viewRange.start + (this.mouseEndX / this.stateController.zoomFactor);
-                if (start !== end) {
+                if (start !== end && this.unitController.viewRangeLength > 1) {
+                    let newViewStart = Math.max(Math.min(start, end), this.unitController.viewRange.start);
+                    let newViewEnd = Math.min(Math.max(start, end), this.unitController.viewRange.end);
+                    if (newViewEnd - newViewStart < 1) {
+                        const center = (newViewStart + newViewEnd) / 2;
+                        newViewStart = center - 0.5;
+                        newViewEnd = center + 0.5;
+                    }
                     this.unitController.viewRange = {
-                        start: Math.max(Math.min(start, end), this.unitController.viewRange.start),
-                        end: Math.min(Math.max(start, end), this.unitController.viewRange.end)
+                        start: newViewStart,
+                        end: newViewEnd
                     }
                 }
                 this.stage.cursor = 'default';


timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -272,7 +272,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                 this.mouseEndX = e.offsetX;
                 const start = this.mouseZoomingStart;
                 const end = this.unitController.viewRange.start + (this.mouseEndX / this.stateController.zoomFactor);
-                if (start !== end) {
+                if (Math.trunc(start) !== Math.trunc(end)) {
```
PatrickTasse, line 275: I think this could happen if the range is too close but on both sides of an integer. I would suggest to use Math.round() instead.

However, if the current range is > 1 ns, and we zoom into this condition, I don't think we should ignore it, I think that instead we should make it so that if rounded start and end are the same, set end = start + 1. So for the user it would still zoom but to a 1 ns window range.

Also, I found that the freezing is due to time-graph-axis-scale.ts, at line 94, because the stepLength is so small, iLo and iHi are very large numbers, beyond integer number precision and in the for loop, the i value is unaffected by the i++ so that loop is infinite.

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -272,12 +272,18 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
-                if (start !== end) {
-                    this.unitController.viewRange = {
-                        start: Math.max(Math.min(start, end), this.unitController.viewRange.start),
-                        end: Math.min(Math.max(start, end), this.unitController.viewRange.end)
-                    }
+                
+                let newViewStart = Math.max(Math.min(start, end), this.unitController.viewRange.start);
+                let newViewEnd = Math.min(Math.max(start, end), this.unitController.viewRange.end);
+                if (newViewEnd - newViewStart < 1) {
+                    newViewStart = Math.trunc(Math.min(newViewStart, newViewEnd));
```
PatrickTasse, line 279: I think from above newViewStart is guaranteed to be the smallest of the two right? So the min is unnecessary.

I would consider adding trunc() on newViewStart and ceil() on newViewEnd above ('let' statements), so that they are always integers, what do you think?

Come to think of it, would that make the < 1 check unnecessary? Since if start and end are different, their trunc() and ceil() cannot possibly be the same integer?

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -272,12 +272,18 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                 this.mouseEndX = e.offsetX;
                 const start = this.mouseZoomingStart;
                 const end = this.unitController.viewRange.start + (this.mouseEndX / this.stateController.zoomFactor);
-                if (start !== end) {
```
PatrickTasse, line 275: I think we need to keep this check, if the user just right-clicks without moving, we want to ignore it and not zoom to 1ns.

As a user (but because I know the implementation) sometimes I abort my zooming by going back to the start pixel...

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
muddana-satish: Updated. Thanks.

PatrickTasse (approved):

bhufmann (approved):

PR #142, opened 2021-06-15
Support opacity for states and symbol annotations

PatrickTasse:

### DIFF
--- a/timeline-chart/src/components/time-graph-annotation.ts
+++ b/timeline-chart/src/components/time-graph-annotation.ts
@@ -11,8 +11,8 @@ export interface TimeGraphAnnotationStyle extends TimeGraphComponentOptions {
     symbol?: string
     size?: number
     color?: number
+    opacity?: number
     verticalAlign?: string
-    opacity?:number
 }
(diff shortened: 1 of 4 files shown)


bhufmann (approved):
LGTM. Thanks!

timeline-chart/src/components/time-graph-annotation.ts
```diff
@@ -11,8 +11,8 @@ export interface TimeGraphAnnotationStyle extends TimeGraphComponentOptions {
     symbol?: string
     size?: number
     color?: number
+    opacity?: number
     verticalAlign?: string
-    opacity?:number
```
MatthewKhouzam, line 15: No point in this, right?

MatthewKhouzam (approved):

timeline-chart/src/components/time-graph-annotation.ts
PatrickTasse: There was missing space so I took the opportunity to move opacity next to color at the same time, for clarity ;)

PR #144, opened 2021-07-01
Press esc to cancel mouse zooming

muddana-satish:

### DIFF
--- a/doc/documentation.md
+++ b/doc/documentation.md
@@ -40,6 +40,7 @@ A reusable time axis component, that can be used independently of the other comp
 * Horizontal zooming can be performed using Ctrl+mouse wheel. The zooming is centered on the mouse cursor position.
 * Horizontal panning can be performed using the middle mouse button or Ctrl+left mouse button.
 * Horizontal zooming selection can be performed using the right mouse button.
+* Horizontal zooming can be cancelled by pressing the `esc` button while zooming using the right click + drag.
 * The view is connected to a time controller instance and synchronizes its viewport, zoom level cursors bi-directionally.
 
 ## Data Model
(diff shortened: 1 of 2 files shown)


MatthewKhouzam (approved):
No Objections

bhufmann:
@muddana-satish  when is this needed?  Any steps to reproduce.
Is this to cancel a long lasting zoom operation 
or 
is there a case where the mouseZooming flag is not reset and pressing esc fixes this condition. If yes then we should figure out why the flag is not reset
or 
any other reasons?

Please let me know.

muddana-satish:
> @muddana-satish when is this needed? Any steps to reproduce.

This is just a UX feature. If the user wants to zoom in by right click drag and then for some reason decides to cancel the zooming, there is no way he can do it. Pressing `esc` can takes him out of the zooming mode. 

In the screencast below, I cancelled zooming mode by pressing `esc`.
Kapture 2021-07-06 at 11 04 34

bhufmann:
> > @muddana-satish when is this needed? Any steps to reproduce.
> 
> This is just a UX feature. If the user wants to zoom in by right click drag and then for some reason decides to cancel the zooming, there is no way he can do it. Pressing `esc` can takes him out of the zooming mode.
> 
> In the screencast below, I cancelled zooming mode by pressing `esc`.

Thanks for the information. Makes sense. Could you please add a sentence to the documentation (under the doc directory) to explain that behaviour? There is a description of the right click drag and it would fit there.

bhufmann (changes requested):
Please add a line in the documentation for that:
https://github.com/theia-ide/timeline-chart/blob/master/doc/documentation.md

muddana-satish:
> Please add a line in the documentation for that:
> https://github.com/theia-ide/timeline-chart/blob/master/doc/documentation.md

Update the documentation.

bhufmann (approved):
LGTM. Thanks!

PR #147, opened 2021-07-09
Fix Mouse zooming overlay

### LINKED ISSUE #145: Mouse zooming doesn't show overlay when scrolled down
Kapture 2021-07-08 at 14 24 45

muddana-satish:
fixes #145

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -348,7 +348,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                     x: mouseStartX,
                     y: 0
                 },
-                height: this.stateController.canvasDisplayHeight,
+                height: Math.max(this.stateController.canvasDisplayHeight, this.rowController.totalHeight),
                 width: this.mouseEndX - mouseStartX
             });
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
lgtm

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -346,7 +346,7 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                 opacity: 0.2,
                 position: {
                     x: mouseStartX,
-                    y: 0
+                    y: this.rowController.verticalOffset
```
PatrickTasse, line 349: If you mouse-wheel up/down while right-click zooming, you can see a gap on top of original vertical offset.

I believe the proper fix is to change height to :

Math.max(this.stateController.canvasDisplayHeight, this.rowController.totalHeight)

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
muddana-satish: Updated. Thanks.

PatrickTasse (approved):

Issue #156, opened 2021-09-21
Incorrect throttling prevents fetching correct time graph data


PatrickTasse:
In TimeGraphChart's view range change handler, new time graph data should sometimes be fetched for the new view range. But if data is already being fetched, the throttling mechanism will just omit the fetching of new data.

Therefore when changing the view range rapidly, for example when zooming with mouse wheel or panning with the mouse, only the first view range change will fetch data, and the subsequent and last view range will not. The end result is time graph data that does not match the latest view range.

Either the data should be fetched for the latest upon completion of the current fetching, or a debouncing function could be used to wait until view range has finished updating to fetch the data for the latest stable view range.

PatrickTasse:
Fixed with https://github.com/theia-ide/timeline-chart/pull/160

PR #160, opened 2021-10-08
Use debounce on view range change

PatrickTasse:
Instead of skipping fetching of new data when it is already in progress,
use debounce to fetch new data after a 400ms delay without range change.

Fixes Issue #156

Change-Id: Icb74e223bebdd4d43ec2f4add81df38e74702393


IbrahimFradj (approved):
LGTM, thanks

PR #162, opened 2021-10-08
Make time graph state label optional

PatrickTasse:
Not all time graph states have a label. Make the variable optional
instead of requiring users to put an empty string.

Change-Id: I980917cfb0d6c761e78905bc5bfef5ee3480da86


bhufmann (approved):
LGTM. Thanks!

PR #169, opened 2021-11-17
Support multi-click events

hriday-panchasara:
Enabling support for double clicking a state
- Ideal use case is to be able to zoom in to double click
- Double click mouse interaction on an element is communicated with timegraph-output-component in trace-extension for custom integrations

Implementation
- PixiJS does not support double click interactions, the recommended approach is to make use of a click event and a timeout: https://github.com/pixijs/pixijs/issues/5910
- Using a counter to check if an element has been clicked, followed by a timeout to confirm a double-click within the specified interval
-  Other left-click interactions such as toggling between elements, clicking rows and time-range selection would not trigger double click
- I've added some commenting in the source code to explain my approach. I can remove them when everything looks good

Example:
- Width of element occupies the viewRange when double-clicked
Screen Recording 2021-11-17 at 3 43 02 PM

Relates to https://github.com/theia-ide/theia-trace-extension/pull/564

Signed-off-by: hriday-panchasara <hriday.panchasara@ericsson.com>

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -8,10 +8,10 @@ import { TimelineChart } from "../time-graph-model";
 import { TimeGraphRowController } from "../time-graph-row-controller";
 import { TimeGraphChartLayer } from "./time-graph-chart-layer";
 import { BIMath } from "../bigint-utils";
-import { debounce } from 'lodash';
+import { debounce, cloneDeep, DebouncedFunc } from 'lodash';
 
 export interface TimeGraphMouseInteractions {
-    click?: (el: TimeGraphComponent<any>, ev: PIXI.InteractionEvent) => void
+    click?: (el: TimeGraphComponent<any>, ev: PIXI.InteractionEvent, clickCount: number) => void
     mouseover?: (el: TimeGraphComponent<any>, ev: PIXI.InteractionEvent) => void
     mouseout?: (el: TimeGraphComponent<any>, ev: PIXI.InteractionEvent) => void
     mousedown?: (el: TimeGraphComponent<any>, ev: PIXI.InteractionEvent) => void
@@ -69,6 +69,13 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     private _keyUpHandler: { (event: KeyboardEvent): void; (event: Event): void; };
     private _mouseWheelHandler: { (ev: WheelEvent): void; (event: Event): void; (event: Event): void; };
     private _contextMenuHandler: { (e: MouseEvent): void; (event: Event): void; };
+
+    // Keep track of the most recently clicked point.
+    // If clicked again during _multiClickTime duration (milliseconds) record multi-click
+    private _recentlyClickedGlobal: PIXI.Point | null = null;
+    private _multiClickTime: number = 500;
+    private _mouseClicks = 0;
+    private _multiClickTimer: DebouncedFunc<() => void>;
 
     constructor(id: string,
         protected providers: TimeGraphChartProviders,
@@ -614,12 +621,37 @@ export class TimeGraphChart extends TimeGraphChartLayer {
 
     protected addElementInteractions(el: TimeGraphComponent<any>) {
         el.displayObject.interactive = true;
+
+        var self = this;
+        this._multiClickTimer = debounce(function(){
+            self._mouseClicks = 0;
+            self._recentlyClickedGlobal = null;
+        }, this._multiClickTime);
+
         el.displayObject.on('click', ((e: PIXI.InteractionEvent) => {
             if (el instanceof TimeGraphStateComponent && !this.mousePanning && !this.mouseZooming) {
                 this.selectState(el.model);
             }
+
+            // Mouse clicks count keeps increasing without limit as long as we keep clicking on the same coordinate.
+            if (this._recentlyClickedGlobal && (this._recentlyClickedGlobal.equals(e.data.global))){
+                this._mouseClicks++;
+            } else {
+                // Only clear the timer and reset the click count if the global position is NOT 
+                // the same one as click 1.
+                this._multiClickTimer.cancel();
+                this._mouseClicks = 1;
+
+                // Store the global position on first click
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
+            }
+
+            // We can use a debouncer to reset the count when no click occurs for a certain period.
+            this._multiClickTimer();
+
+            // Click callback includes count parameter to record subsequent clicks on the same point
             if (this.mouseInteractions && this.mouseInteractions.click) {
-                this.mouseInteractions.click(el, e);
+                this.mouseInteractions.click(el, e, this._mouseClicks);
             }
         }).bind(this));
         el.displayObject.on('mouseover', ((e: PIXI.InteractionEvent) => {


example/src/index.ts
```diff
@@ -161,6 +161,9 @@ timeGraphChart.registerMouseInteractions({
     click: el => {
         console.log('click: ' + el.constructor.name, el.model);
     },
+    doubleClick: el => {
+        console.log('click: ' + el.constructor.name, el.model);
```
PatrickTasse, line 165: 'doubleClick'

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -617,10 +628,39 @@ export class TimeGraphChart extends TimeGraphChartLayer {
         el.displayObject.on('click', ((e: PIXI.InteractionEvent) => {
             if (el instanceof TimeGraphStateComponent && !this.mousePanning && !this.mouseZooming) {
                 this.selectState(el.model);
+                // Time range selection using left-click should not trigger a double click action
```
PatrickTasse, line 631: This if-clause is for states only because it can call selectState().

But I believe that the doubleClick handling should apply for all kinds of elements (states, arrows, annotations, etc.)

PatrickTasse: I am surprised that we get a PIXI click event when you mouseDown, mouseMove, then mouseUp (although, the mouseDown and mouseUp need to be on the same element for the click to be called).

I would propose that our click/doubleClick mouseInterationEvent be triggered only when the mouse does not move between mouseDown and mouseUp. So store the coordinates of mouseDown and it must be the same in the click event otherwise we do not propagate it. This will prevent click on range selection, in a simpler way. I would also avoid storing things related to clicks in the element itself.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -617,10 +628,39 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                if (!(this.unitController.selectionRange && (this.unitController.selectionRange.start !== this.unitController.selectionRange.end))) {
+                    if (el.clickStatus){
+                        // perform double click action
+                        // changing event type from 'click' to 'doubleClick'
+                        e.type = 'doubleClick';
+                        // reset counter
+                        el.clickStatus = false;
+                    }else{
```
PatrickTasse, line 639: spaces before and after all { and }

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -617,10 +628,39 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                            if (el.clickStatus){
+                                el.clickStatus = false; 
+                            }
+                        }, this._dbClickTime);          
+                    }
+                }
             }
-            if (this.mouseInteractions && this.mouseInteractions.click) {
+
+            if (e.type === 'click' && this.mouseInteractions && this.mouseInteractions.click) {
```
PatrickTasse, line 658: Looking at this for example: https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

It seems like on the doubleClick we should also see a click event just before. e.g. click(1), click(2), doubleClick.

When I triple-, quadruple-click and more, it seems that doubleClick is sporadically sent, and it depends on the timing. There should be only one doubleClick, e.g. click(1), click(2), doubleClick, click(3), click(4), click(5), ...

I would propose that this class just keeps track locally of the count. The count keeps increasing without limit as long as we keep clicking on the same coordinate. We can use a debouncer to reset the count when no click occurs for a certain period. Then the doubleClick event would be fired only when count == 2.

I would also propose to improve the click callback to include the count parameter (unless we can somehow include the count in the event, but not sure if good idea or even possible to modify it since we didn't create it).

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
hriday-panchasara: Thanks for the feedback Patrick! The implementation looks much cleaner now. Let me know if you need any additional changes

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+            this._mouseClicks++;
+
+            if (this._mouseClicks === 1) {
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
+                this._dblClickTimer = setTimeout(function(){
+                    self._mouseClicks = 0;
+                    self._recentlyClickedGlobal = null;
+                }, this._dbClickTime);
+            } else if (this._recentlyClickedGlobal && (this._recentlyClickedGlobal.equals(e.data.global))){
+                clearTimeout(this._dblClickTimer);
```
PatrickTasse, line 640: I don't think we should clear on doubleClick, otherwise you get a doubleClick event on click(4), click(6), etc.

I think we should use a debouncer so every subsequent click on the same point (if close enough to previous click, not to first click...) will extend the timer. It's only when the debouce timer expires that we reset the click count and position. Or also if we get any click at a different position.

Minor point, if done like that then the dbClickTime constant should be renamed, as it's a timer for all multiple clicks (2,3,4,etc.)

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                e.type = 'doubleClick';
+                this._mouseClicks = 0;
+                this._recentlyClickedGlobal = null;
+            }
+
+            if (e.type === 'click' && this.mouseInteractions && this.mouseInteractions.click) {
                 this.mouseInteractions.click(el, e);
+            } else if (e.type === 'doubleClick' && this.mouseInteractions && this.mouseInteractions.doubleClick){
```
PatrickTasse, line 648: As mentioned, should not be an if-else, the doubleClick event should be in addition to the click event.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                this.mouseInteractions.doubleClick(el,e);
```
PatrickTasse, line 649: space after ,

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+
+            if (this._mouseClicks === 1) {
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
+                this._dblClickTimer = setTimeout(function(){
+                    self._mouseClicks = 0;
+                    self._recentlyClickedGlobal = null;
+                }, this._dbClickTime);
+            } else if (this._recentlyClickedGlobal && (this._recentlyClickedGlobal.equals(e.data.global))){
+                clearTimeout(this._dblClickTimer);
+                e.type = 'doubleClick';
```
PatrickTasse, line 641: Just checking, is it OK to put (our own) type value that is not in the PIXI InteractionEvent.type API?

Be careful, if we send both click and doubleClick as proposed, the type should be ´click' in the click. Maybe better to use the count as condition to detect doubleClick?

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+            var self = this;
+            this._mouseClicks++;
+
+            if (this._mouseClicks === 1) {
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
+                this._dblClickTimer = setTimeout(function(){
+                    self._mouseClicks = 0;
+                    self._recentlyClickedGlobal = null;
+                }, this._dbClickTime);
+            } else if (this._recentlyClickedGlobal && (this._recentlyClickedGlobal.equals(e.data.global))){
```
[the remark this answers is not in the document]
hriday-panchasara: The else-if helps ensure that click events are not fired on doubleClick, so subsequent clicks on the same point would look like this:
click(1)
doubleClick(2)
click(3)
doubleClick(4)

In this the desired behavior? I thought we need to record doubleClick on click(4), click(6)..

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +626,27 @@ export class TimeGraphChart extends TimeGraphChartLayer {
             if (el instanceof TimeGraphStateComponent && !this.mousePanning && !this.mouseZooming) {
                 this.selectState(el.model);
             }
-            if (this.mouseInteractions && this.mouseInteractions.click) {
+            
+            var self = this;
+            this._mouseClicks++;
+
+            if (this._mouseClicks === 1) {
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
```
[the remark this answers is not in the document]
hriday-panchasara: E.g. 1) Time selection through mouse dragging calls the following events:
mouseover -> mousedown -> mouseup -> click

 E.g. 2) Time selection using shift + left mouse click calls the following events:
mouseover -> mousedown -> mouseup -> click -> mousedown -> mouseup -> click

Click events are always sent at the end of time selection, but they would not trigger doubleClick since 'e.data.global' would capture different positions as cursor positions have changed during drag. But I feel that in  E.g. 1 (mouse dragging) we should prevent click event being triggered

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +625,26 @@ export class TimeGraphChart extends TimeGraphChartLayer {
             if (el instanceof TimeGraphStateComponent && !this.mousePanning && !this.mouseZooming) {
                 this.selectState(el.model);
             }
-            if (this.mouseInteractions && this.mouseInteractions.click) {
-                this.mouseInteractions.click(el, e);
+            
+            var clickCount = 1;
```
PatrickTasse, line 629: I think this variable is not necessary, you can just use _mouseClicks?

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +625,26 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+            var self = this;
+            this._mouseClicks++;
+
+            if (this._mouseClicks === 1) {
+                this._recentlyClickedGlobal = cloneDeep(e.data.global);
+                this._dblClickTimer = setTimeout(function(){
```
PatrickTasse, line 635: I would suggest to use debounce(). This way, you can keep clicking fast and it won't go back to 1 in the middle of the multi-click.

Note, the debounce should be called for all clicks not just the first one. So the only thing special to do on first click is to store the global position.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +625,26 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                    self._mouseClicks = 0;
+                    self._recentlyClickedGlobal = null;
+                }, this._dbClickTime);
+            } else if (this._recentlyClickedGlobal && (this._recentlyClickedGlobal.equals(e.data.global))){
```
PatrickTasse, line 639: I would suggest that we only clear the timer and reset the click count if the global position is NOT the same one as click 1. In that case we should store the new position and restart another timer. Maybe this can be checked before the previous if-case so that you can reuse the code for first click. (click on a different point is then seen as a new first click, even if done quickly).

Then the 2nd, 3rd, etc. clicks on the same global position will just fall-through and send the event with the correct click count. No special case for count==2, it should just continue 2, 3, 4, ...

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +625,26 @@ export class TimeGraphChart extends TimeGraphChartLayer {
...
+                clearTimeout(this._dblClickTimer);
+                clickCount = 2;
+                this._mouseClicks = 0;
+                this._recentlyClickedGlobal = null;
+            }
+
+            if (e.type === 'click' && this.mouseInteractions && this.mouseInteractions.click) {
```
PatrickTasse, line 646: Now that we have clickCount instead of doubleClick, the e.type will always be 'click', that condition can be removed.

example/src/index.ts
```diff
@@ -161,9 +161,6 @@ timeGraphChart.registerMouseInteractions({
     click: el => {
         console.log('click: ' + el.constructor.name, el.model);
     },
-    doubleClick: el => {
-        console.log('doubleClick: ' + el.constructor.name, el.model);
-    },
```
PatrickTasse: Please just amend these changes in the first commit. I don't think we need two separate commits?

PatrickTasse (changes requested):

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -70,6 +70,13 @@ export class TimeGraphChart extends TimeGraphChartLayer {
     private _mouseWheelHandler: { (ev: WheelEvent): void; (event: Event): void; (event: Event): void; };
     private _contextMenuHandler: { (e: MouseEvent): void; (event: Event): void; };
 
+    // Keep track of the most recently clicked point.
+    // If clicked again during _dbClickTime duration (milliseconds) trigger double click
+    private _recentlyClickedGlobal: PIXI.Point | null = null;
+    private _dbClickTime: number = 500;
+    private _mouseClicks = 0;
+    private _dblClickTimer: DebouncedFunc<() => void>;
```
PatrickTasse, line 78: Maybe these should be called _multiClickTime/_multiClickTimer now that it's more than just doubleClick?

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -618,8 +625,32 @@ export class TimeGraphChart extends TimeGraphChartLayer {
             if (el instanceof TimeGraphStateComponent && !this.mousePanning && !this.mouseZooming) {
                 this.selectState(el.model);
             }
+            
+            var self = this;
+            this._dblClickTimer = debounce(function(){
+                self._mouseClicks = 0;
+                self._recentlyClickedGlobal = null;
+            }, this._dbClickTime);
```
PatrickTasse, line 633: It should not create a new instance of the debounce function every time. Move this declaration to the beginning of the add method, outside of the click event handler.

PatrickTasse (changes requested):

PatrickTasse (approved):
Thanks for the contribution!

bhufmann (approved):
Looks good to me. I tested it with the theia-trace-extension patch https://github.com/theia-ide/theia-trace-extension/pull/564

PR #171, opened 2021-12-14
Unit tests

Rodrigoplp-work:
Adding Jest and some unit tests to components.

Tests can be executed from the root of the project with:

```
yarn test
```

Or from `./timeline-chart`, where the test outputs are formatted in a
more readable way:

```
cd timeline-chart
yarn test --verbose --watch
```

Current coverage:

/timeline-chart/src: 16.3%
/timeline-chart/src/components: 23.4%
/timeline-chart/src/layer: 0%

Signed-off-by: Rodrigo Pinto <rodrigo.pinto@calian.ca>

https://user-images.githubusercontent.com/88502808/146078921-e74e084a-f8cb-46e1-bebe-4dd7ad85c671.mov


PatrickTasse (approved):

MatthewKhouzam (approved):
Ok, I approve, but this is only testing up to the display part. We cannot tell with this if the render is bad. And trust me, openGL is a can of worms.

Issue #54, opened 2020-10-08
Gridlines are not on rounded time boundaries


PatrickTasse:
The time graph gridlines and axis ticks are on rounded multiples of the step length away from the start time of the view range.
But the absolute time of the gridline will not be on a rounded number because the time graph does not seem to be aware of the time offset of its view range, which is only known by the (external) implementation of the numberTranslator.

The time offset should be available and used by the time graph so that it can set the gridlines and axis ticks on rounded boundaries.

bhufmann:
Can you  provide a screenshot for illustration purposes?

PatrickTasse:
Fixed with https://github.com/theia-ide/timeline-chart/pull/172

PR #172, opened 2021-12-16
Make gridlines aligned to rounded numbers

PatrickTasse:
Add the offset to the time graph unit controller.

Use the offset to make sure that gridlines are set on rounded absolute
times, not rounded relative times to the start of the absolute range.

Calculate the maximum label width using the view range end time instead
of the view range length.

Fixes issue #54

Change-Id: I5ff00c57ae2859703c8eeb8b4db88bc1655d985f

### DIFF
--- a/timeline-chart/src/components/time-graph-axis-scale.ts
+++ b/timeline-chart/src/components/time-graph-axis-scale.ts
@@ -68,21 +68,20 @@ export class TimeGraphAxisScale extends TimeGraphComponent<null> {
         if (this.unitController.viewRangeLength > 0) {
             let labelWidth = 0;
             if (this.unitController.numberTranslator) {
-                const label = this.unitController.numberTranslator(this.unitController.viewRangeLength);
+                const label = this.unitController.numberTranslator(this.unitController.viewRange.end);
                 if (label) {
                     const style = new PIXI.TextStyle({ fontSize: 10 });
                     const textMetrics = PIXI.TextMetrics.measureText(label, style);
                     labelWidth = textMetrics.width;
                 }
             }
-            const stepLength = this.getStepLength(labelWidth);
+            const stepLength = BigInt(this.getStepLength(labelWidth));
             const canvasDisplayWidth = this.stateController.canvasDisplayWidth;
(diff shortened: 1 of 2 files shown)


timeline-chart/src/components/time-graph-axis-scale.ts
```diff
@@ -91,7 +90,7 @@ export class TimeGraphAxisScale extends TimeGraphComponent<null> {
                     };
                     let label;
                     if (drawLabels && this.unitController.numberTranslator) {
-                        label = this.unitController.numberTranslator(time);
+                        label = this.unitController.numberTranslator(time - this.unitController.offset);
```
bhufmann, line 93: in the numberTranslator of the theia-trace-extension the offset is subtracted from the value passed. I haven't run the code, but it doesn't feel right to subtract it here too.

bhufmann (commented):
@PatrickTasse I added a question

timeline-chart/src/components/time-graph-axis-scale.ts
PatrickTasse: In trace-context-component.tsx the start time (offset) is added to the provided number.

bhufmann: I see, the variable `time` here is now the "real" time and not the offset time as it was in the previous code. Makes sense.

bhufmann (changes requested):
The example is broken.

image

It works without this change.

Rodrigoplp-work:
I couldn't reproduce the failure Bernd found. The example worked when I run the branch locally. Tested in Chrome, Firefox and Safari.

Rodrigoplp-work (approved):

PatrickTasse:
I could reproduce, it was the offset being undefined in the unit controller. I added an initialization to BigInt(0).

bhufmann:
> I couldn't reproduce the failure Bernd found. The example worked when I run the branch locally. Tested in Chrome, Firefox and Safari.

It happened when right away when running it with GitPod.

bhufmann (approved):
Looks good to me. Thanks!

PR #173, opened 2022-01-07
Update Webpack in example project to version 5

Rodrigoplp-work:
| Package | Current | Updated to |
| --- | --- | --- |
| webpack | 4.22.0 | 5.65.0 |
| webpack-cli | 3.1.2 | 4.9.1 |
| webpack-dev-server | 3.1.10 | 4.7.2 |

The updates resolve Dependabot alerts for the following packages:

- glob-parent;
- ansi-html;
- ansi-regex.

As of now there are also alerts for `json-schema` and `trim-off-newlines`,
but those are dependencies of `lerna` 4.0.0, which does not have an
update yet.

Signed-off-by: Rodrigo Pinto <rodrigo.pinto@calian.ca>

### DIFF
--- a/example/webpack.config.js
+++ b/example/webpack.config.js
@@ -5,8 +5,8 @@ module.exports = {
     entry: ['./src/index.ts'],
     devtool: "source-map",
     devServer: {
-        contentBase: path.join(__dirname, 'public'),
-        disableHostCheck: true,
+        static: path.join(__dirname, 'public'),
+        allowedHosts: 'all',
         hot: true
     },
     module: {
--- a/example/package.json
+++ b/example/package.json
@@ -6,17 +6,17 @@
   "scripts": {
     "build": "webpack --mode=development",
     "watch": "webpack --watch --mode=development",
-    "start": "webpack-dev-server --host 0.0.0.0 --mode=development"
+    "start": "webpack serve --host 0.0.0.0 --mode=development"
   },
   "devDependencies": {
     "@types/lodash.throttle": "^4.1.4",
     "@types/node": "^8.0.26",
     "file-loader": "^2.0.0",
     "source-map-loader": "^0.2.4",
-    "ts-loader": "^5.2.2",
-    "webpack": "^4.22.0",
-    "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.1.10"
+    "ts-loader": "9.2.6",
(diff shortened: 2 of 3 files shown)


bhufmann (changes requested):
Running the example in Gitpod doesn't work anymore. I get the following message:
**Invalid Host header**

Could you please look into this?

BTW, running it locally on my computer it works.

Rodrigoplp-work:
How curious! The "invalid host header" is an error that was fixed by the changes in `example/webpack.config.js`. This attempt on Gitpod went well:

https://user-images.githubusercontent.com/88502808/148972798-5e4be4a2-6624-4cbc-8500-770bfe6105dd.mov

bhufmann:
> How curious! The "invalid host header" is an error that was fixed by the changes in `example/webpack.config.js`. This attempt on Gitpod went well:
> 
I tried it again (with firefox and chrome) and I consistently get the same error. Not sure how to trouble-shoot this. @Rodrigoplp-work any ideas?

PatrickTasse (approved):

example/webpack.config.js
```diff
@@ -5,8 +5,10 @@ module.exports = {
     entry: ['./src/index.ts'],
     devtool: "source-map",
     devServer: {
-        contentBase: path.join(__dirname, 'public'),
-        disableHostCheck: true,
+        static: path.join(__dirname, 'public'),
+        allowedHosts: [
```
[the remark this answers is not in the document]
Rodrigoplp-work: You are right, Bernd. Gitpod doesn't run on localhost, and that is why the `allowedHosts: [0.0.0.0]` would throw an error.
The video I recorded above worked because in there I was running Gitpod on the master branch, not on this PR's branch.
Now I changed `allowedHosts` to `all`.
Also, while investigating I saw that the `ts-loader` package was throwing exceptions with Webpack 5, so I updated it from version 5.2.2 to 9.2.6.

dist/main.js
```diff
@@ -0,0 +1,14 @@
+/*
```
bhufmann, line 1: @Rodrigoplp-work  what's the reason for adding this file? Please let me know.

Rodrigoplp-work: Removed.

bhufmann (approved):
Works now. Thanks!

PR #175, opened 2022-02-04
Update data model on resize

PatrickTasse:
When the time graph canvas width is increased by a resize event, the
data model may no longer be at the appropriate resolution. Invoke the
debounced method to maybe fetch new data, in the same way as is done
when the view range is changed.

Update the data model when the resolution is changed either way (not
only when zooming in) so that an appropriately sized and consistent data
model is always used.

Change-Id: I33373a3fa490ae94e157e20638423de149c91021


bhufmann (approved):
Makes sense. Looks good to me. Thanks.

Rodrigoplp-work (approved):

PR #180, opened 2022-02-18
Ignore duplicate data provider requests and discard obsolete responses

### LINKED ISSUE #179: Unnecessary duplicate data provider requests
In TimeGraphChart:maybeFetchNewData() method, a new request is discarded if it the current data model already corresponds to the new request.

But if a previous request for the same view range is ongoing and awaiting response, the new request will be allowed to proceed, causing two concurrent requests for the same data model.

Furthermore, since requests are asynchronous, there is no guarantee on the order of reception of their responses, and if the older requests comes back last, it will overwrite the data model from the latest request with an obsolete data model.

A request for the same view range as an ongoing request should be ignored, and a response for a model that is not for the latest request should be discarded.

PatrickTasse:
Fixes #179

Change-Id: I5e0a96738f27f9191338f6d647f3cff8ad15aab7


bhufmann (approved):
Looks good to me. Thanks!

Rodrigoplp-work (approved):

PR #189, opened 2022-04-04
Destroy TimeGraphLayer in one operation

### LINKED ISSUE #187: Closing time graph chart can be very slow
Closing a time graph chart that has a very large number of components (states, annotations, arrows) can take a very long time, > 20 seconds when there are > 120k elements. This is because the time graph layer destroys each of its children one by one.

It should be checked if the whole layer can be destroyed in a single operation.

PatrickTasse:
Add a child container to TimeGraphLayer. Add children to the child
container instead of the layer container.

Destroy the layer and its children graphics objects in one call instead
of destroying each child component one-by-one.

When removing all children, destroy the child container in one
operation, destroying all children graphics objects.

The TimeGraphComponent.destroy() method is no longer called recursively
on the child components when destroying the layer or removing all
children. Components are not expected to have any local resources to
release at destruction, other than the graphics object.

Make TimeGraphComponent.destroy() also destroy all children graphics
objects of the component in one operation. Remove the now unnecessary
override of destroy() in TimeGraphArrowComponent.

In TimeGraphNavigatorHandle, add document listener in the mouse down
handler, and remove the document listener in the mouse up handler.

Fixes https://github.com/eclipse-cdt-cloud/timeline-chart/issues/187.

### DIFF
--- a/timeline-chart/src/layer/time-graph-layer.ts
+++ b/timeline-chart/src/layer/time-graph-layer.ts
@@ -13,10 +13,11 @@ export abstract class TimeGraphLayer {
     protected children: TimeGraphComponent<any>[];
     protected stage: PIXI.Container;
     protected layer: PIXI.Container;
+    private childContainer: PIXI.Container | undefined;
 
     constructor(protected id: string) {
         this.children = [];
-        this.layer = new PIXI.Container;
+        this.layer = new PIXI.Container();
     }
 
     protected addChild(child: TimeGraphComponent<any>, parent?: TimeGraphParentComponent) {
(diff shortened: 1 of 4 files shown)


timeline-chart/src/layer/time-graph-layer.ts
```diff
@@ -73,7 +73,8 @@ export abstract class TimeGraphLayer {
     protected afterAddToContainer() { }
 
     destroy() {
-        this.removeChildren();
+        this.layer.destroy();
```
bhufmann, line 76: removeChildren() calls destroy() on each child component which destroys the corresponding display object and WebGL resource. Not calling destroy() would cause WebGL resources not being disposed.

bhufmann (changes requested):

timeline-chart/src/layer/time-graph-layer.ts
PatrickTasse: Now calling destroy({children: true}) to destroy all WebGL resources of the layer and its children components recursively.

If there are any user-application resources to release in any layer it can be done in the layer's destroy() method.

With the changes, there are no more user-application resources to release in any of the components. If it's needed in the future I do not think it should be done through the component's destroy() method.

bhufmann (approved):
Code looks good to me and the destroying of the layer is much faster now.

Rodrigoplp-work (approved):

PR #199, opened 2022-05-02
Page scrolling on vertical and horizontal scrollbars

williamsyang-work:
Implements page scrolling on horizontal and vertical scrollbars.  The user can right-click to move the navigator one page in the correct direction.

screen-recorder-mon-may-02-2022-14-44-17

Updates comments and variable names on previous snap-on-click feature implementation for clarity.

Note: If testing in the timeline-chart example, right click will trigger a context menu.  This will not happen in the trace-viewer-extension, since the default is prevented onContextMenu.  You can disable context menu by opening the browser console and typing `window.oncontextmenu = e => e.preventDefault();`

### DIFF
--- a/timeline-chart/src/layer/time-graph-vertical-scrollbar.ts
+++ b/timeline-chart/src/layer/time-graph-vertical-scrollbar.ts
@@ -111,14 +111,19 @@ export class TimeGraphVerticalScrollbarBackground extends TimeGraphComponent<nul
     constructor(protected rowController: TimeGraphRowController, protected stateController: TimeGraphStateController, protected factor: number) {
         super("vscroll_background");
         this.addEvent("mousedown", event => {
-            let yPosition = event.data.getLocalPosition(this._displayObject).y;
-            let vOffset = (yPosition/this.stateController.canvasDisplayHeight) * this.rowController.totalHeight;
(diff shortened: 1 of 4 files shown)


PatrickTasse:
In the timeline-chart example, the context-menu is disabled when right-clicking on the graph itself. I think it should also be disabled when right-clicking on the horizontal scroll bar, even if it's disabled from theia-trace-extension.

timeline-chart/src/layer/time-graph-vertical-scrollbar.ts
```diff
@@ -9,6 +9,9 @@ export class TimeGraphVerticalScrollbar extends TimeGraphChartLayer {
     protected navigatorHandle: TimeGraphVerticalScrollbarHandle;
     protected navigatorBackground: TimeGraphVerticalScrollbarBackground;
     protected selectionRange?: TimeGraphRectangle;
+    private _contextMenuHandler: { (e: MouseEvent): void; (event: Event): void; };
+    private __verticalOffsetChangedHandler: () => void;
```
PatrickTasse, line 13: Is double underscore intentional?

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. Thanks.

PR #200, opened 2022-05-11
Fix missing row updates

PatrickTasse:
When the update flag in maybeFetchNewData() is set to true, it means all
rows must be refreshed (e.g. annotations might have changed). It was
only refreshing the visible rows, and when the other rows later became
visible they were considered valid and not refreshed.

Mark all existing rows as needing to be refreshed by setting their
providedModel object to undefined.


bhufmann (approved):
Thanks. This fixes the issue described in PR comment:  https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/740#pullrequestreview-969573245

Rodrigoplp-work (approved):

PR #205, opened 2022-06-07
Support coarse and fine update of time graph data

PatrickTasse:
To enable this feature, provide a coarse resolution factor that is
higher than 1 to the TimeGraphChart constructor.

When a coarse resolution factor is provided, the first update of time
graph data is made using the coarse resolution. Once this update is
completed, a debounced fine resolution second update is triggered.

The resolution check against provided model is changed to require better
resolution, to prevent a coarse model from being fetched and used to
overwrite a previously fetched fine model.

When no row needs update at coarse resolution, try to update with fine
resolution. This allows the update to trigger when the time graph canvas
width is increased.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -36,6 +36,7 @@ export const keyBoardNavs: Record<string, Array<string>> = {
 export type TimeGraphRowStyleHook = (row: TimelineChart.TimeGraphRowModel) => TimeGraphRowStyle | undefined;
 
 const VISIBLE_ROW_BUFFER = 3; // number of buffer rows above and below visible range
+const FINE_RESOLUTION_FACTOR = 1; // fine resolution factor or default to disable coarse update
 
 export class TimeGraphChart extends TimeGraphChartLayer {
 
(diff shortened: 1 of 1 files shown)


marco-miller (approved):
A co-approval or review would be needed; no rush intended though. I wasn't able to visually perceive the effect of changing this new constructor parameter value to (e.g.) `8`, but I am new to more subtle `timeline-chart` behaviours. Hence the need for a second review or more trials.

I was able to hit all the touched breakpoints though, without perceiving any issue. I used a `context-switches` (kernel+ust traces) experiment with the Thread Status view, in `theia-trace-extension`. Debugged in VS Code using Chrome naturally.

The logic LGTM, albeit having to manually change the code to exercise a different behaviour. Under-the-hood performance while fetching model data is the likely benefit, although not felt when I tried it. I also don't understand all the touched code flows or branches as we speak.

timeline-chart/src/layer/time-graph-chart.ts
```diff
@@ -487,7 +492,13 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                     this.ongoingRequest = undefined;
                 }
                 this.isNavigating = false;
+                if (this._coarseResolutionFactor !== FINE_RESOLUTION_FACTOR) {
```
PatrickTasse, line 495: I realized that if this is already the 2nd (fine) update, we will unnecessarily debounce a 3rd fine update. Although when it executes, it is likely to have no rows to update and abort. But we still should not attempt it if it was already the fine update.

bhufmann (approved):
Looks good to me.

Issue #203, opened 2022-05-30
Location of Time axis ticks should be configurable


bhufmann:
Right now the Time axis tick lines location is on top if the tick time value. This is ok when the time axis is on the bottom of view graph. However, for the case that the time axis is drawn on top of the graph, then the ticks should be below the time value strings, which is used in the theia-trace-extension example (https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/755). 

![image](https://user-images.githubusercontent.com/413424/171049111-f24b63ef-f7e7-45c2-82b1-88691fd6e9c3.png)

The location of the tick lines should be configurable so that it can be drawn on top or below the time strings.

PR #206, opened 2022-06-08
Add support for configurable location of time axis ticks

hoangphamEclipse:
This commit adds support for configurable location of time axis ticks. Currently, the ticks can only be displayed above the time labels in the time axis, which do not make sense UI-wise when we have graphs that are displayed under the time axis. This commit allows the ticks of the time axis to be displayed dynamically either on top or under the time labels by adding a configuration parameter to the TimeGraphAxisScale component, so that developers can position the ticks correctly.

Resolve [Issue 203](https://github.com/eclipse-cdt-cloud/timeline-chart/issues/203) of the Theia Trace Extension. This also relates to [Issue 755](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/755) of the Theia Trace Extension.

### DIFF
--- a/timeline-chart/src/layer/time-graph-axis.ts
+++ b/timeline-chart/src/layer/time-graph-axis.ts
@@ -17,17 +17,21 @@ export class TimeGraphAxis extends TimeGraphLayer {
     private _keyUpHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
     private _keyDownHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
 
-    constructor(id: string, protected style?: { color?: number, lineColor?: number }) {
+    constructor(id: string, protected style?: { color?: number, lineColor?: number, verticalAlign?: string }) {
         super(id);
     }
 
     protected getOptions(): TimeGraphAxisStyle {
         let color;
         let lineColor;
+        let verticalAlign: string | undefined = 'top'; // Default position is top, same as CSS verticalAlign is 'baseline'
+        
         if (this.style) {
             color = this.style.color;
             lineColor = this.style.lineColor;
+            verticalAlign = this.style.verticalAlign;
         }
+
         return {
             height: this.stateController.canvasDisplayHeight,
             width: this.stateController.canvasDisplayWidth,
@@ -36,7 +40,8 @@ export class TimeGraphAxis extends TimeGraphLayer {
                 y: 0
             },
             color,
-            lineColor
+            lineColor,
+            verticalAlign
         }
     }
 
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -132,7 +132,7 @@ const timeGraphAxisContainer = new TimeGraphContainer({
 }, unitController, axisCanvas);
 axisHTMLContainer.appendChild(timeGraphAxisContainer.canvas);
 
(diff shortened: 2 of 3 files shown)


bhufmann (changes requested):
It looks good and it works well. I tested it with the theia-trace-extension. I have one suggestion, please change the vertical alignment of the example to bottom (see index.ts). With both configurations are available in the example and axis test.

timeline-chart/src/layer/time-graph-axis.ts
```diff
@@ -17,17 +17,21 @@ export class TimeGraphAxis extends TimeGraphLayer {
     private _keyUpHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
     private _keyDownHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
 
-    constructor(id: string, protected style?: { color?: number, lineColor?: number }) {
+    constructor(id: string, protected style?: { color?: number, lineColor?: number, verticalLinePosition?: VerticalLinePosition }) {
```
PatrickTasse, line 20: To me it seems odd that the external user of TimeGraphAxis needs to import the type used in the constructor from an internal component. Seems like it should be defined in this class.

PatrickTasse: What I would suggest is to have it as top?: boolean.

So by default, the axis is at the bottom (like an normal chart).

Then, this layer will use the boolean to set the verticalLinePosition.
top=true -> VerticalLinePosition.BOTTOM
else -> VerticalLinePosition.TOP

bhufmann:
@hoangphamEclipse is this PR ready for review? I'm asking because it's marked as draft.

hoangphamEclipse:
@bhufmann Yes it is ready for review. Sorry I forget to revert it back to Ready for review.

timeline-chart/src/layer/time-graph-axis.ts
```diff
@@ -17,17 +17,21 @@ export class TimeGraphAxis extends TimeGraphLayer {
     private _keyUpHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
     private _keyDownHandler: { (event: KeyboardEvent): void; (this: Document, ev: KeyboardEvent): any; };
 
-    constructor(id: string, protected style?: { color?: number, lineColor?: number }) {
+    constructor(id: string, protected style?: { color?: number, lineColor?: number, isTop?: boolean }) {
```
PatrickTasse, line 20: Sorry, we had further discussion and we would suggest that we use the vertical-align string property from CSS (verticalAlign in code):

https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

So if you align to the top, that means you stick to the top of your layout cell, then the ticks should be on top.
If you align to the bottom the ticks should be at the bottom.

Looking at the link above, it seems their default is 'baseline' and in the example that aligns to the top. So perhaps we could default to 'top'.

<pre>
+---|---|---|---+
|   1   2   3   |  verticalAlign = 'top'
|               |
+---------------+

+---------------+
|               |
|   1   2   3   |  verticalAlign = 'bottom'
+---|---|---|---+
</pre>

PatrickTasse (changes requested):

PatrickTasse (approved):

bhufmann (approved):
Looks good! Thanks

PR #208, opened 2022-07-06
Prevent Idle Renders

### LINKED ISSUE #7: Reduce GPU load
Even without any interaction, the Chrome profiler shows more than 50% load on the GPU. 
Can we improve on that?

williamsyang-work:
All PIXI.Applications are initiated on the PIXI shared ticker, allowing them to be controlled simultaneously.

PIXI.Ticker.shared is controlled via TimeGraphRenderController and RenderEvents.  RenderEvents fires custom events that are handled by event handlers initiated in the RenderController.

RenderController is created in the UnitController; this is only because we need one RenderController and there is already only one UnitController.

RenderEvents currently fires start-render events in:
- TimeGraphComponent on creation and update.
- StateController when scrolling up/down

Component creation/update handles almost all required rendering events, except scrolling up and down.  This is because there is no vertical scrollbar from timeline-chart being used in theia-trace-extension.  (For horizontal scrolling, the updating of the scrollbar-handle component will fire a render event).

Stop render is debounced at a 1500 timeout after start render.

Screenshot at 2022-07-06 15-50-22

Fixes #7

### DIFF
--- a/timeline-chart/src/time-graph-render-controller.ts
+++ b/timeline-chart/src/time-graph-render-controller.ts
@@ -1,0 +1,65 @@
+import { throttle, debounce } from 'lodash';
+import * as PIXI from 'pixi.js-legacy';
+
+const START_RENDER_STRING = 'startPixiRender';
+const STOP_RENDER_STRING = 'stopPixiRender';
+
+const throttledStart = throttle(() => {
+    window.dispatchEvent(new Event(START_RENDER_STRING));
+}, 450, { leading: true });
+
+const debouncedStop = debounce(() => {
+    window.dispatchEvent(new Event(STOP_RENDER_STRING));
+}, 1000);
+
+interface RenderEvents {
+    startRender: () => void;
+    stopRender: () => void;
+};
+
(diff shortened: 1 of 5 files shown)


timeline-chart/src/time-graph-render-controller.ts
```diff
@@ -0,0 +1,58 @@
...
+        throttledStart();
+    },
+    /**
+     * Fires an event that will stop the PIXI.Ticker.
+     * This will stop rendering.
+     * StopRender event handler is debounced @ 1.5s.
+     * Event handler located in time-graph-render-controller.ts
+     */
+    stopRender: () => {
+        window.dispatchEvent(new Event(STOP_RENDER_STRING));
```
PatrickTasse, line 31: For consistency, this could be a debouncedStop() method that dispatches the event, and the event listener then would stop the ticker immediately.

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. I run the chrome performance analyzer and it works as described. Thanks for the improvement.

PR #213, opened 2022-09-11
Viewport implementation

williamsyang-work:
# Viewport Implementation for Timeline-Chart

## Purpose & Concept
Previously, the canvas/PIXI.Stage was being re-drawn before being re-rendered every time the view changed.  Re-draws are expensive and should be only used when necessary.  

This change implements the concept of having a "world".  The world can be drawn, and changes things like panning / zooming / resizing can be abstracted using the PIXI library. Modifying PIXI variables like position and scale can be used to modify what is shown instead of clearing the canvas of all PIXI objects and re-drawing them in the correct place.

Current implementation only properly handles horizontal panning, or moving left and right.  Other actions like zooming, window resize, and others, still require a manual re-draw.  These can be implemented in the future.

### UnitController Changes

1. `worldRange` is added.
2. `worldRenderFactor` added.  This determines how many `viewRanges` left and right of the new view to get data for.  It's currently set to 1, which is the same as before.
3. `updateWorldRangeFromViewRange`.  This does what it says, using `WORLD_RENDER_FACTOR` to calculate the new world size to set.  This is called in `TimeGraphChart#maybeFetchNewData` to determine what data to fetch.

### TimeGraphViewportLayer Class

1. This class modifies canvas.position.x to shift the Viewport Layer left and right.  Each canvas' position is modified independently; this is a refactor requested by @bhufmann which significantly reduced the amount of code needed.
2. `getWorldPixel` is an abstraction of `TimeGraphLayer#getPixel`.  It gets the pixel relative to the world start, and has the option to clamp to world bounds.
3. Class inheritance is now `TimeGraphLayer` => `TimeGraphViewportLayer` => `TimeGraphChartLayer`.

### StateController OnZoomChange Implemented and Used

1. `onZoomChanged` handler was not functional or used before.
2. Used because there was an edge case when `worldRange = absoluteRange` and `viewRange < worldRange/absoluteRange`.  The entire trace was rendered to the world, and changing the view range by zooming out did not trigger any `onWorldRender` events that previously triggered  `onViewRangeChanged`.  `onZoomChanged` handler added to `TimeGraphChart` to force a world re-draw when the zoom changes.  (_This also adds a nice touch when zooming out, where the world becomes scaled-down and rendered relative to where it was before_).

## Performance Gains - Horizontal Panning

This is a "back of the napkin" check.  Trace used is "Many-Threads", with a size of ~8MB.
No other views were open, and Resource Status was resized for maximum vertical height.

CPU: AMD Ryzen 7 3700X
GPU:  EVGA GeForce RTX 3070
RAM: Corsair Vengeance 2x64 GB @ 3200 MHz

### Pre-Update
Re-draw + Render resulted in frame times from ~220-450ms, or about 3 FPS.
Memory (blue line near the top) progressively spikes as we perform each re-draw.

image

### Viewport
Frame times reduced to ~30-50ms.  Frames increased to ~30 FPS.
Stable memory curve.

_**Note:** We are significantly more zoomed-in on this picture._

image

## Tech Debt
There is a shortcut solution for ensuring the container background and row lines are always visible.  They are updated every time the view range changes, and don't "beat to the same drum" for rendering/updating as the other Viewport layers.  See `TimeGraphViewportContainer#updateBackground` and `TimeGraphChart#ensureRowLinesFitWidth`.

### DIFF
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -3,10 +3,10 @@ import { TimeGraphChart } from "timeline-chart/lib/layer/time-graph-chart";
 import { TimeGraphUnitController } from "timeline-chart/lib/time-graph-unit-controller";
 import { TimeGraphRowController } from "timeline-chart/lib/time-graph-row-controller";
 import { TimeGraphNavigator } from "timeline-chart/lib/layer/time-graph-navigator";
-import { TimeGraphContainer } from "timeline-chart/lib/time-graph-container";
 import { TimeGraphChartCursors } from "timeline-chart/lib/layer/time-graph-chart-cursors";
 import { TimeGraphChartSelectionRange } from "timeline-chart/lib/layer/time-graph-chart-selection-range";
 import { TimeGraphAxisCursors } from "timeline-chart/lib/layer/time-graph-axis-cursors";
+import { TimeGraphContainer } from "timeline-chart/lib/time-graph-container";
 // import { timeGraph } from "timeline-chart/lib/test-data";
 import { TimelineChart } from "timeline-chart/lib/time-graph-model";
 import { TimeGraphStateStyle } from "timeline-chart/lib/components/time-graph-state";
@@ -36,6 +36,7 @@ container.style.width = styleConfig.mainWidth + "px";
 const testDataProvider = new TestDataProvider(styleConfig.mainWidth);
 let timeGraph = testDataProvider.getData({});
 const unitController = new TimeGraphUnitController(timeGraph.totalLength);
+unitController.worldRenderFactor = 3;
 unitController.numberTranslator = (theNumber: bigint) => {
     let num = theNumber.toString();
     if (num.length > 6) {
@@ -54,11 +55,7 @@ const providers = {
         };
     },
     dataProvider: (range: TimelineChart.TimeGraphRange, resolution: number) => {
-        const length = range.end - range.start;
-        const overlap = length * BigInt(10);
-        const start = range.start - overlap > BigInt(0) ? range.start - overlap : BigInt(0);
-        const end = range.end + overlap < unitController.absoluteRange ? range.end + overlap : unitController.absoluteRange;
-        const newRange: TimelineChart.TimeGraphRange = { start, end };
+        const newRange: TimelineChart.TimeGraphRange = range;
         const newResolution: number = resolution * 0.1;
         timeGraph = testDataProvider.getData({ range: newRange, resolution: newResolution });
         return {
@@ -132,8 +129,9 @@ const timeGraphAxisContainer = new TimeGraphContainer({
 }, unitController, axisCanvas);
 axisHTMLContainer.appendChild(timeGraphAxisContainer.canvas);
 
+const timeAxisCursors = new TimeGraphAxisCursors('timeGraphAxisCursors', { color: styleConfig.cursorColor });
 const timeAxisLayer = new TimeGraphAxis('timeGraphAxis', { color: styleConfig.naviBackgroundColor, verticalAlign: 'bottom'});
-timeGraphAxisContainer.addLayers([timeAxisLayer]);
+timeGraphAxisContainer.addLayers([timeAxisLayer, timeAxisCursors]);
 
 const chartHTMLContainer = document.createElement('div');
 chartHTMLContainer.id = 'main_chart';
@@ -153,13 +151,12 @@ chartHTMLContainer.appendChild(timeGraphChartContainer.canvas);
 const timeGraphChartGridLayer = new TimeGraphChartGrid('timeGraphGrid', rowHeight);
 const timeGraphChart = new TimeGraphChart('timeGraphChart', providers, rowController);
 const timeGraphChartArrows = new TimeGraphChartArrows('timeGraphChartArrows', rowController);
-const timeAxisCursors = new TimeGraphAxisCursors('timeGraphAxisCursors', { color: styleConfig.cursorColor });
 const timeGraphSelectionRange = new TimeGraphChartSelectionRange('chart-selection-range', { color: styleConfig.cursorColor });
 const timeGraphChartCursors = new TimeGraphChartCursors('chart-cursors', timeGraphChart, rowController, { color: styleConfig.cursorColor });
 const timeGraphChartRangeEvents = new TimeGraphRangeEventsLayer('timeGraphChartRangeEvents', providers);
 
 timeGraphChartContainer.addLayers([timeGraphChartGridLayer, timeGraphChart,
-    timeGraphChartArrows, timeAxisCursors, timeGraphSelectionRange,
+    timeGraphChartArrows, timeGraphSelectionRange,
     timeGraphChartCursors, timeGraphChartRangeEvents]);
 
 timeGraphChart.registerMouseInteractions({
--- a/timeline-chart/src/layer/time-graph-chart-arrows.ts
+++ b/timeline-chart/src/layer/time-graph-chart-arrows.ts
@@ -7,11 +7,11 @@ export class TimeGraphChartArrows extends TimeGraphChartLayer {
 
     protected arrows: Map<TimelineChart.TimeGraphArrow, TimeGraphArrowComponent>;
     protected rowIds: number[] = [];
-    private _updateHandler: { (): void; (viewRange: TimelineChart.TimeGraphRange): void; (viewRange: TimelineChart.TimeGraphRange): void; };
+    private _updateHandler: { (): void; (worldRange: TimelineChart.TimeGraphRange): void; (worldRange: TimelineChart.TimeGraphRange): void; };
 
     protected afterAddToContainer() {
         this._updateHandler = (): void => this.update();
-        this.unitController.onViewRangeChanged(this._updateHandler);
+        this.stateController.onWorldRender(this._updateHandler);
 
         this.rowController.onVerticalOffsetChangedHandler(verticalOffset => {
             this.layer.position.y = -verticalOffset;
@@ -24,13 +24,12 @@ export class TimeGraphChartArrows extends TimeGraphChartLayer {
         if (sourceIndex === -1 || destinationIndex === -1) {
             return undefined;
         }
-        const relativeStartPosition = arrow.range.start - this.unitController.viewRange.start;
         const start: TimeGraphElementPosition = {
-            x: this.getPixel(relativeStartPosition),
+            x: this.getWorldPixel(arrow.range.start),
             y: (sourceIndex * this.rowController.rowHeight) + (this.rowController.rowHeight / 2)
         }
         const end: TimeGraphElementPosition = {
-            x: this.getPixel(relativeStartPosition + arrow.range.end - arrow.range.start),
+            x: this.getWorldPixel(arrow.range.end),
             y: (destinationIndex * this.rowController.rowHeight) + (this.rowController.rowHeight / 2)
         }
         return { start, end };
@@ -83,7 +82,7 @@ export class TimeGraphChartArrows extends TimeGraphChartLayer {
 
     destroy() : void {
         if (this.unitController) {
-            this.unitController.removeViewRangeChangedHandler(this._updateHandler);
+            this.stateController.removeWorldRenderHandler(this._updateHandler);
         }
         super.destroy();
     }
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -144,7 +144,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
             }
         };
         this.onCanvasEvent('mousedown', this._mouseDownHandler);
-        this.unitController.onViewRangeChanged(this._updateHandler);
+        this.stateController.onWorldRender(this._updateHandler);
         this.unitController.onSelectionRangeChange(this._updateHandler);
         this.update();
     }
@@ -224,8 +224,8 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     update() {
         if (this.unitController.selectionRange) {
-            const firstCursorPosition = this.getPixel(this.unitController.selectionRange.start - this.unitController.viewRange.start);
-            const secondCursorPosition = this.getPixel(this.unitController.selectionRange.end - this.unitController.viewRange.start);
+            const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
+            const secondCursorPosition = this.getWorldPixel(this.unitController.selectionRange.end);
             const firstCursorOptions = {
                 color: this.color,
                 height: this.stateController.canvasDisplayHeight,
@@ -268,7 +268,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     destroy() : void {
         if (this.unitController) {
-            this.unitController.removeViewRangeChangedHandler(this._updateHandler);
+            this.stateController.removeWorldRenderHandler(this._updateHandler);
             this.unitController.removeSelectionRangeChangedHandler(this._updateHandler);
         }
         if (this._mouseDownHandler) {
--- a/timeline-chart/src/layer/time-graph-chart-selection-range.ts
+++ b/timeline-chart/src/layer/time-graph-chart-selection-range.ts
@@ -1,11 +1,10 @@
 import { TimeGraphRectangle } from "../components/time-graph-rectangle";
 import { TimelineChart } from "../time-graph-model";
-import { TimeGraphLayer } from "./time-graph-layer";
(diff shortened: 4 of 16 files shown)


williamsyang-work:
https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/832#pullrequestreview-1118972354

**Problem:** When increasing height, we only fetch and re-render the newly displayed rows.  We were always resetting the world range with every request.  This led to the new rows to be in the correct place, but the old rows became skewed because they weren't updated from the previous world range.

**Solution:** Only update the world range if we're updating all of the rows.  Added a quick check and changed the conditional for updating the world range.

timeline-chart/src/layer/time-graph-chart-selection-range.ts
```diff
@@ -17,8 +16,8 @@ export class TimeGraphChartSelectionRange extends TimeGraphLayer {
 
     protected updateScaleAndPosition() {
         if (this.unitController.selectionRange && this.selectionRange) {
-            const firstCursorPosition = this.getPixel(this.unitController.selectionRange.start - this.unitController.viewRange.start);
-            const width = this.getPixel(this.unitController.selectionRange.end - this.unitController.selectionRange.start)
+            const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
```
bhufmann, line 19: This only works if the parent container is an instance of TimeGraphViewportContainer and not a just TimeGraphContainer. Shouldn't APIs be defined in TimeGraphContainer and here you use these APIs. TimeGraphViewportContainer should just overwrite the relevant methods, but in this class only calls an API doesn't have care where it's implemented (TimeGraphContainer or TimeGraphViewportContainer)

timeline-chart/src/layer/time-graph-chart-arrows.ts
```diff
@@ -24,13 +24,12 @@ export class TimeGraphChartArrows extends TimeGraphChartLayer {
         if (sourceIndex === -1 || destinationIndex === -1) {
             return undefined;
         }
-        const relativeStartPosition = arrow.range.start - this.unitController.viewRange.start;
         const start: TimeGraphElementPosition = {
-            x: this.getPixel(relativeStartPosition),
+            x: this.getWorldPixel(arrow.range.start),
```
bhufmann, line 28: same comment as in timeline-chart/src/layer/time-graph-chart-selection-range.ts about APIs.

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -224,8 +224,8 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     update() {
         if (this.unitController.selectionRange) {
-            const firstCursorPosition = this.getPixel(this.unitController.selectionRange.start - this.unitController.viewRange.start);
-            const secondCursorPosition = this.getPixel(this.unitController.selectionRange.end - this.unitController.viewRange.start);
+            const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
```
bhufmann, line 227: same comment as in timeline-chart/src/layer/time-graph-chart-selection-range.ts about APIs.

timeline-chart/src/layer/time-graph-range-events-layer.ts
```diff
@@ -1,27 +1,27 @@
...
-        this._viewRangeUpdateHandler = (): void => this.update();
-        this.unitController.onViewRangeChanged(this._viewRangeUpdateHandler);
+        this._worldRangeUpdateHandler = (): void => this.update();
+        this.viewport.onWorldRender(this._worldRangeUpdateHandler);
     }
 
     protected addRangeEvent(rangeEvent: TimelineChart.TimeGraphAnnotation) {
-        const start = this.getPixel(rangeEvent.range.start - this.unitController.viewRange.start);
-        const end = this.getPixel(rangeEvent.range.end - this.unitController.viewRange.start);
+        const start = this.getWorldPixel(rangeEvent.range.start);
```
bhufmann, line 23: same comment as in timeline-chart/src/layer/time-graph-chart-selection-range.ts about APIs.

bhufmann (changes requested):
I have few things to look into. 

See also comment here: https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/832#discussion_r986086848 that shows an problem when disabling the viewport

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -2,8 +2,18 @@ import { TimelineChart } from "./time-graph-model";
...
+
+    
     protected viewRangeChangedHandlers: ((newRange: TimelineChart.TimeGraphRange) => void)[];
     protected _viewRange: TimelineChart.TimeGraphRange;
+    
+    /**
+     * How many extra view ranges left and right of view range.
+     * Increasing this increases world size / data cache.
+     */
+    public WORLD_RENDER_FACTOR = BigInt(1);
```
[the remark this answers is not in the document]
williamsyang-work: I thought about this.  I'm not sure of a clean way to use a float for the WORLD_RENDER_FACTOR, but I'm sure I can work something out.  I'll work on this.

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -45,6 +63,24 @@ export class TimeGraphUnitController {
...
     }
 
+    /**
+     * Handles changes on world range change, but before the states/rows render.
+     * If the behavior looks weird  or glitchy after implementation and you're working in 
+     * a TimeGraphViewportLayer / TimeGraphChartLayer class, consider using:
+     * 
+     * TimeGraphChartLayer#onWorldRangeChanged
+     */
+    onWorldRangeChanged(handler: (worldRange: TimelineChart.TimeGraphRange) => void) {
```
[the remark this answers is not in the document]
williamsyang-work: I thought this would be needed, but everything worked best with TimeGraphViewportContainer#handleOnWorldRender vs onWorldRangeChange.  I can remove the handlers, but there might be a use for it in the future. UnitController.onZoomChange was not used before this, but there was code there to handle it (sort of).

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -56,6 +92,14 @@ export class TimeGraphUnitController {
         }
     }
 
+    updateWorldRangeFromViewRange() {
```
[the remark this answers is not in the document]
williamsyang-work: The world range doesn't necessarily change with the view range.  For example, if we are panning right, but are still within the bounds of the "world", the world range does not change despite the view range changing.  The only time the world range changes is when a redraw is called in TimeGraphChart#maybeFetchNewData.  From there, we redraw the world based on the current viewRange.

timeline-chart/src/time-graph-viewport-container.ts
```diff
@@ -0,0 +1,205 @@
...
+    destroy = () => {
+        this.unitController.removeViewRangeChangedHandler(this._onViewRangeChanged);
+        this.removeWorldRenderHandler(this._onWorldRender);  // Why not.
+        super.destroy();
+    }
+
+
+    // Debugging functions
+    // These can be useful for future feature implementations. 
+    debug = (): any => {
```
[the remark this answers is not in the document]
williamsyang-work: They can be removed.  I'd like to keep them because they can be helpful for future work, but ultimately this is up to you.  I can also save this code somewhere else and put it back in if needed for future work.

timeline-chart/src/layer/time-graph-chart-selection-range.ts
williamsyang-work: I'm not sure I understand this one.

TimeGraphViewportLayer#getWorldPixel is just an abstraction of getPixel.  I noticed for every call of #getPixel we were doing: `TimeWeWantPixelFor - unitController.viewRange.start`.  (This became `TimeWeWantPixelFor - unitController.worldRange.start` for the viewport implementation).  getWorldPixel's only purpose is to make the code cleaner, so it can be removed without consequence. 

Furthermore:

All layers added to a TimeGraphViewportContainer must be a TimeGraphViewportLayer.  Abstract class TimeGraphChartLayer now extends TimeGraphViewportLayer, and all layers that are added to the ChartContainer and MarkersContainer (timegraph-output-component.tsx @ 504-505) are now ViewportLayers with access to the inherited method.

There is a check at TimeGraphViewportContainer#addLayer @ 29.  If a layer is not a ViewportLayer, an error is thrown.

The only non-viewport layers are the top-axis, vertical, and horizontal scrollbars.

If the design is wrong, or this doesn't answer your question, please let me know.

bhufmann: This is comment is related to the comment in the related `theia-trace-extension` PR [1]. It's not clear from the API when to use a TimeGraphViewportLayer and when to use TimeGraphChartLayer, as well that they cannot be mixed. 

Maybe all layers should be viewport layer and the TimeGraphChartLayer is just an internal base class. Or is there a different way? WDYT?

Btw, the documentation [2] should also have a description about the TimeGraphViewportLayer and its usage.

[1] https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/832#discussion_r986086848
[2] https://github.com/eclipse-cdt-cloud/timeline-chart/blob/master/doc/documentation.md

timeline-chart/src/time-graph-viewport-container.ts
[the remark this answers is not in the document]
williamsyang-work: These have been removed.

timeline-chart/src/layer/time-graph-chart-selection-range.ts
williamsyang-work: The need to use a viewport container has been removed.

timeline-chart/src/layer/time-graph-chart-arrows.ts
williamsyang-work: This should also be resolved.

timeline-chart/src/layer/time-graph-chart-cursors.ts
williamsyang-work: This should also be resolved.

timeline-chart/src/layer/time-graph-range-events-layer.ts
williamsyang-work: This should also be resolved.

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -1,9 +1,21 @@
...
+
+    
     protected viewRangeChangedHandlers: ((newRange: TimelineChart.TimeGraphRange) => void)[];
     protected _viewRange: TimelineChart.TimeGraphRange;
+    
+    /**
+     * How many extra view ranges left and right of view range.
+     * Increasing this increases world size / data cache.
+     */
+    private _worldRenderFactor = 1.5;
```
bhufmann, line 15: the default value of 1.5 leads to a cache of 1.5 times the view range left and right. Shouldn't the default be 1.0?

bhufmann (changes requested):
When testing, the issue that was reported https://github.com/eclipse-cdt-cloud/timeline-chart/pull/213#issuecomment-1261413574 and fixed is back.

Please rebase the patches to latest master.

timeline-chart/src/time-graph-unit-controller.ts
williamsyang-work: Set to 1.

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -45,6 +65,32 @@ export class TimeGraphUnitController {
...
+
+    removeWorldRangeChangedHandler(handler: (worldRange: TimelineChart.TimeGraphRange) => void) {
+        const index = this.worldRangeChangedHandlers.indexOf(handler);
+        if (index > -1) {
+            this.worldRangeChangedHandlers.splice(index, 1);
+        }
+    }
+
+    updateWorldRangeFromViewRange() {
+        const { start: vs, end: ve } = this.viewRange;
```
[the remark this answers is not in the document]
williamsyang-work: Fixed.

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -1,9 +1,21 @@
...
 
 export class TimeGraphUnitController {
+
+    
     protected viewRangeChangedHandlers: ((newRange: TimelineChart.TimeGraphRange) => void)[];
     protected _viewRange: TimelineChart.TimeGraphRange;
+    
+    /**
+     * How many extra view ranges left and right of view range.
+     * Increasing this increases world size / data cache.
```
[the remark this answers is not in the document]
williamsyang-work: Fixed.

timeline-chart/src/time-graph-state-controller.ts
```diff
@@ -106,4 +122,23 @@ export class TimeGraphStateController {
         this.handlePositionChange();
     }
 
+    /**
+     * Performs actions that should be handled after the world is rendered.
+     * Some actions can be performed before we render the data, but some need to be performed post-render.
+     */
```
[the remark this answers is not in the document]
williamsyang-work: I just removed the comment.  If you want me to put in a better one LMK.

williamsyang-work:
> When testing, the issue that was reported [#213 (comment)](https://github.com/eclipse-cdt-cloud/timeline-chart/pull/213#issuecomment-1261413574) and fixed is back.

I tested this with the new changes and it's working as intended for me.  Can you post a screenshot?

bhufmann:
> > When testing, the issue that was reported [#213 (comment)](https://github.com/eclipse-cdt-cloud/timeline-chart/pull/213#issuecomment-1261413574) and fixed is back.
> 
> I tested this with the new changes and it's working as intended for me. Can you post a screenshot?
I checked out the latest code, but I can't reproduce anymore. It seems to be fine now.

example/src/index.ts
```diff
@@ -4,6 +4,7 @@ import { TimeGraphUnitController } from "timeline-chart/lib/time-graph-unit-cont
 import { TimeGraphRowController } from "timeline-chart/lib/time-graph-row-controller";
 import { TimeGraphNavigator } from "timeline-chart/lib/layer/time-graph-navigator";
 import { TimeGraphContainer } from "timeline-chart/lib/time-graph-container";
+import { TimeGraphViewportContainer } from "timeline-chart/lib/time-graph-viewport-container";
```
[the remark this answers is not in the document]
bhufmann: This hasn't been addressed.

example/src/index.ts
```diff
@@ -36,6 +37,7 @@ container.style.width = styleConfig.mainWidth + "px";
 const testDataProvider = new TestDataProvider(styleConfig.mainWidth);
 let timeGraph = testDataProvider.getData({});
 const unitController = new TimeGraphUnitController(timeGraph.totalLength);
+unitController.WORLD_RENDER_FACTOR = BigInt(3);
```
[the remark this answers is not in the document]
bhufmann: This hasn't been addressed

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -45,6 +65,32 @@ export class TimeGraphUnitController {
         }
     }
 
+    /**
+     * Handles changes on world range change, but before the states/rows render.
+     * If the behavior looks weird  or glitchy after implementation and you're working in
```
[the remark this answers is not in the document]
bhufmann: Comment still needs to be updated.

bhufmann (changes requested):
I tested it again and it looks good. A few things still need to be addressed:
- remaining comments
- compilation error in index.ts
- rebase to latest master

If these things are addressed you can squash it.

bhufmann (approved):
Looks good to me know. I'll ask for someone to do another pass. After that you need to squash the commits and then we can merge. Thanks a lot for this improvement.

timeline-chart/src/time-graph-unit-controller.ts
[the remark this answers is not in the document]
Rodrigoplp-work: Apparently not in use. I agree with Bernd, it should be removed even if may be of use in the future.

timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -45,6 +66,24 @@ export class TimeGraphUnitController {
         }
     }
 
+    onWorldRangeChanged(handler: (worldRange: TimelineChart.TimeGraphRange) => void) {
+        this.worldRangeChangedHandlers.push(handler);
+    }
+
+    removeWorldRangeChangedHandler(handler: (worldRange: TimelineChart.TimeGraphRange) => void) {
```
Rodrigoplp-work, line 73: Apparently not in use. It should be removed.

Rodrigoplp-work (approved):

bhufmann (approved):
It looks good to me too. Please squash the commits then we can merge it.

PR #216, opened 2022-11-11
Markers refactor

williamsyang-work:
### Navigate markers by pressing < and >

Adds this functionality and exposes the method.  This method is used in theia-trace-extension to be called on key press.


timeline-chart/src/components/time-graph-marker.ts
```diff
@@ -0,0 +1,25 @@
...
+        displayObject?: PIXI.Graphics
+    ) {
+        super(id, _model, xStart, xEnd, _row, _style, displayWidth, displayObject);
+    }
+
+    public get model(): TimelineChart.TimeGraphMarkerState {
+        return this._model;
+    }
+
+}
```
bhufmann, line 25: no new line at the end

timeline-chart/src/layer/time-graph-markers-chart.ts
```diff
@@ -0,0 +1,122 @@
...
+export class TimeGraphMarkersChart extends TimeGraphChart {
+
+    private _selectedMarker?: TimelineChart.TimeGraphMarkerState;
+
+    constructor(id: string, protected providers: TimeGraphChartProviders, protected rowController: TimeGraphRowController) {
+        super(id, providers, rowController);
+        
+        this.registerMouseInteractions({
+            click: el => {
+                if (el instanceof TimeGraphMarkerState) {
```
bhufmann, line 22: I'm not able to select any marker state. I set a breakpoint but it never hit. Not sure what the reason is.

timeline-chart/src/layer/time-graph-markers-chart.ts
```diff
@@ -0,0 +1,122 @@
...
+            }
+        });
+    }
+
+    protected createNewState = (markerModel: TimelineChart.TimeGraphMarkerState, rowComponent: TimeGraphRow): TimeGraphMarkerState | undefined => {
+        const { xStart, xEnd, displayWidth, elementStyle } = this.getNewStateParams(markerModel);
+        return new TimeGraphMarkerState(markerModel.id, markerModel, xStart, xEnd, rowComponent, elementStyle, displayWidth);
+    }
+
+    onKeydown = (event: KeyboardEvent) => {
```
bhufmann, line 40: navigation of states are done in time-graph-chart-cursors.ts and done differently. Based on the selected row (y-axis) and time (x-axis) it finds the next state. For markers and in time-graph-chart-cursors.ts, you could loop through all markers for the row and time range and move the cursor to the time of the next marker available. 

An advantage is that you don't need to store next & prev markers in each marker after fetching the marker data which is unnecessary processing. On demand searching for the next marker by looping through all markers of a row should be acceptable.

timeline-chart/src/time-graph-model.ts
```diff
@@ -40,6 +40,13 @@ export namespace TimelineChart {
         readonly data?: { [key: string]: any }
     }
 
+    export interface TimeGraphMarkerState extends TimeGraphState {
```
bhufmann, line 43: The disadvantage of storing that you need to store next & prev markers in each marker after fetching the marker data which is unnecessary processing. Unnecessary, because users won't navigate through most of them, only some of them. On demand searching for the next marker by looping through all markers of a row should be acceptable.

bhufmann (changes requested):
Thanks for the updates. Some general and initial comments.

bhufmann (commented):
One more thing, navigating of view markers that spawn the whole vertical view should be considered as well, probably in a separate PR. Having next/previous on that would allow users to move from e.g. frame to frame, if the view marker represent a certain periodic time frame.

We will also need to allow for configuration of marker categories to navigate between. For example, sometime we want to go from frame to frame, other time from row marker to row marker. This is another increment of the feature.

timeline-chart/src/layer/time-graph-markers-chart.ts
williamsyang-work: I'm not sure; this is working for me.  Maybe try again when I push my changes.

timeline-chart/src/layer/time-graph-markers-chart.ts
williamsyang-work: I have implemented this change.  It searches by stateId instead of time, which I think is simpler.

timeline-chart/src/time-graph-model.ts
williamsyang-work: On demand searching implemented.

bhufmann (changes requested):
Thanks for the contribution. I like being able to navigate the through the marker axis. Also, navigating per marker row is great.

I noticed that when no marker is selected then the navigation is not applied. You can reproduce it with a lost events. I think, when a marker row is selected, then navigating left should selected the search should start from the current timestamp (cursor) like for states. WDYT?

I had to update the theia trace extension to use it.  It would be good that the example app of this repo can demonstrate this feature as well.

timeline-chart/src/layer/time-graph-markers-chart.ts
bhufmann: I was trying to select a symbol marker. I realized that this feature is for the marker row. I'm able to select a marker now.

williamsyang-work:
> Thanks for the contribution. I like being able to navigate the through the marker axis. Also, navigating per marker row is great.
> 
> I noticed that when no marker is selected then the navigation is not applied. You can reproduce it with a lost events. I think, when a marker row is selected, then navigating left should selected the search should start from the current timestamp (cursor) like for states. WDYT?

I think this makes sense, I will work on adding this in

> I had to update the theia trace extension to use it. It would be good that the example app of this repo can demonstrate this feature as well.

I was going to wait for the NPM package to be updated before adding the sync.  Sorry for that!  I'll get another one up.

bhufmann:
> > I had to update the theia trace extension to use it. It would be good that the example app of this repo can demonstrate this feature as well.
> 
> I was going to wait for the NPM package to be updated before adding the sync. Sorry for that! I'll get another one up.

No problem. Any chance to add it the example code of this repo?

williamsyang-work:
Creating a different class for the markers is not cohesive with the current architecture.  TimeGraphMarkersChart and TimeGraphMarkerState are removed, and functionality is implemented in TimeGraphChart.

bhufmann (changes requested):
After the chat yesterday and looking at the code, I think we should try to use the same approach for navigation of states like in the main chart using the cursors. In the end that's what we do we move the cursors. However, I wasn't sure if it would work. So I tried it and uploaded a draft commit https://github.com/eclipse-cdt-cloud/timeline-chart/pull/237 and corresponding patch draft https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/920. 

Could you please try it? There are some TODOs which I won't have time to do. If you could update your PR with my change and fix the TODOs there, that would be good.

bhufmann:
> After the chat yesterday and looking at the code, I think we should try to use the same approach for navigation of states like in the main chart using the cursors. In the end that's what we do we move the cursors. However, I wasn't sure if it would work. So I tried it and uploaded a draft commit #237 and corresponding patch draft [eclipse-cdt-cloud/theia-trace-extension#920](https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/920).
> 
> Could you please try it? There are some TODOs which I won't have time to do. If you could update your PR with my change and fix the TODOs there, that would be good.

@williamsyang-work any progress with this one?

bhufmann:
@williamsyang-work I think you can close this one. #239 provides the navigation feature.

PR #218, opened 2022-11-18
Remove zooming selection display object when fetching new data.

hoangphamEclipse:
To remove the zooming selection layer from the timeline chart, it is necessary to 1) remove the zooming selection object from the graph, and 2) delete the zooming selection object to reset it. Currently, when fetching new data, the code only performs step 2), but not 1). This leads to an issue where the zooming selection layer is not removed when users select a new zooming selection after zooming out. This commit adds step 1) so that when new data is fetched, the zooming selection is drawn and removed properly.

Fixes [Trace Compass Issue 839](https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/839)

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -516,6 +516,10 @@ export class TimeGraphChart extends TimeGraphChartLayer {
                         this.selectStateInNavigation();
                     }
                     if (this.mouseZooming) {
+                        if (this.zoomingSelection) {
+                            this.removeChild(this.zoomingSelection);
+                        }
+
                         delete this.zoomingSelection;
                         this.updateZoomingSelection();
                     }


hoangphamEclipse:
**A question for debate**: Should we try to centralize the code that delete the zooming selection layer? Having similar code for similar purposes in multiple place might make debugging harder in the future.  We can either move the code to a separate function [\*], and have both the functions below call [\*] (My preferred method). Or, we can try moving the code from the first function into the second function. If so, should I fix it with this PR too (as this is more of a refactor change, but small)?

https://github.com/eclipse-cdt-cloud/timeline-chart/blob/3b44886a700e95e77441fd6e29e1cfe8cf82b6bd/timeline-chart/src/layer/time-graph-chart.ts#L519-L523

https://github.com/eclipse-cdt-cloud/timeline-chart/blob/3b44886a700e95e77441fd6e29e1cfe8cf82b6bd/timeline-chart/src/layer/time-graph-chart.ts#L379-L406

Rodrigoplp-work:
> **Small question**: Should we try to centralize the code that delete the zooming selection layer? Having similar code for similar purposes in multiple place might make debugging harder in the future. We can either move the code to a separate function [*], and have both the functions below call [*] (My preferred method). Or, we can try moving the code from the first function into the second function. If so, should I fix it with this PR too (as this is more of a refactor change, but small)?
> 
> https://github.com/eclipse-cdt-cloud/timeline-chart/blob/3b44886a700e95e77441fd6e29e1cfe8cf82b6bd/timeline-chart/src/layer/time-graph-chart.ts#L519-L523
> 
> https://github.com/eclipse-cdt-cloud/timeline-chart/blob/3b44886a700e95e77441fd6e29e1cfe8cf82b6bd/timeline-chart/src/layer/time-graph-chart.ts#L379-L406

As of now it looks like just two lines of code. I suggest not to bother putting that in a new function.

Later, when you push the rest of your PR, I'll check again to see if that is indeed the case.

bhufmann (changes requested):
Looks good to me. Thanks!

bhufmann (approved):
Looks good to me. Thanks!

Rodrigoplp-work:
I think I still have non-deleted old selections. I zoomed twice, then panned to the left.

https://user-images.githubusercontent.com/88502808/203643532-afadc4af-47ac-4ee0-9827-efb0933c6a23.mp4

Rodrigoplp-work (changes requested):
I believe I still have non-deleted selections. I zoomed twice and then panned to the left.

https://user-images.githubusercontent.com/88502808/203643808-63511e64-7206-4259-b234-f9eab24e116c.mp4

Rodrigoplp-work:
I run more tests.
The problem doesn't depend on the size of the trace, I could reproduce it even on small traces, but with the overview and some other chart like CPU Usage or the Event Table on.
It seems that after zooming and panning, sometimes an extra zoomingSelection is drawn on the chart, usually off-screen, to the left near the start.
It looks like it happens in `updateZoomingSelection`. It is hard to debug because it doesn't happen when running timeline-chart alone, only when it's linked to theia-trace-extension. So it could be related to how the variables are being updated when there are too many things being draw on screen.

hoangphamEclipse:
I tried to run the resource status view with the overview opened, and 2 other XY views, and 2 event tables. I still cannot reproduce the issue. I added some logs in the `updateZoomingSelection` function. Sometimes, the function seems like it starts drawing a second rectangles right after drawing the first one, without deleting the first one first. However, even when this occurs in the log, I still don't see the rectangle that you got. 

So, @bhufmann, can you take a look at the issue Rodrigo reported, can you replicate this issue?

Also, @Rodrigoplp-work, in the example you showed me, the overview overlay is still orange, so I assume that you are not on master for the trace extension. In the new version, the trace overview zoom selection is supposed to be inverted. Unless you already try this, can you can try testing with the master branch of the trace extension? (Maybe this issue has to do with some other bug fixes).

Rodrigoplp-work (approved):
I used master in theia-trace-extension and I still could reproduce it.
Now, it looks like the problem only happens when I use my Apple mouse (on my linux machine) to select, not the touchpad.
To be fair, the Apple mouse doesn't behave 100% well on linux. Also, it doesn't have 3 buttons, but a single surface that works differently depending on where you click.
So, it could be that the problem I see is because the right button of my mouse is not always being properly recognized.
Being this such a niche use case, I suggest we merge your changes and address the problem I saw in a different ticket in the future, in the case we manage to reproduce it consistently.

hoangphamEclipse:
Thank you Rodrigo. I was suspecting an OS issue too, since all three of us all use different OS for testing.

bhufmann:
I'm fine with the approach to merge this fix. We can backlog the OS specific problem and pick-up later. @hoangphamEclipse go ahead and merge it.

PR #219, opened 2022-11-22
Apply opacity to state label

hriday-panchasara:
If state has opacity apply it to the font color as well

Signed-off-by: hriday-panchasara <hriday.panchasara@ericsson.com>


bhufmann (approved):
Looks good to me. Thanks!

Rodrigoplp-work (approved):

PR #226, opened 2022-12-19
Improve rendering for text labels

### LINKED ISSUE #225: Repeated scrolling with mouse wheel causes UI to freeze due label rendering
When running the timeline-chart with the `theia-trace-extension` and zooming many times with the mouse wheel in a short time periods, causes the UI to be very unresponsive up to a point that it freezes the UI. 

Preliminary investigation shows that the main culprit is the clearing and recreating of the pixi BitmapText causes the issue. Which is executed a lot of times. 

image

hoangphamEclipse:
Currently, when the timeline chart renders text labels for states, it deletes all existing labels, then creates new labels, and adds them to the graph. This is not efficient and makes rendering of the labels sluggish when zooming in and out of the graph. This commit removes the code that delete the text labels objects on every re-render, and updates the text content of the labels instead of re-creating them.

Fixes #225.

### DIFF
--- a/timeline-chart/src/components/time-graph-state.ts
+++ b/timeline-chart/src/components/time-graph-state.ts
@@ -17,6 +17,8 @@ export class TimeGraphStateComponent extends TimeGraphComponent<TimelineChart.Ti
     static fontController: FontController = new FontController();
 
     protected _options: TimeGraphStyledRect;
+    private textLabelObject: PIXI.BitmapText | undefined;
+    private textWidth: number;
 
     constructor(
         id: string,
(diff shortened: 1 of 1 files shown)


bhufmann (approved):
It looks good to me. It improves the overall performance when many labels have to be re-drawn in a short time.

bhufmann:
> Every state has a label, this is a great first step, now can we make every label part of a map in the view. We can look it up at the construction time, knowing our size and text. So "HoangThuanPham" would save labels for "HoangThuanPham", "HoangThuan..." "HoangThua..." "HoangThu..." "HoangTh..." "HoangT..." "Hoang..." etc... then instead of doing a new labelobject at line 76, you make a lookup in the view.
> 
> Just a thought for later.
> 
> Approved

Interesting suggestion for caching BitmapText objects to avoid costly updating of a single BitmapText. I'll create a issue tracker from this comment and we can pick-up later.

PR #229, opened 2022-12-21
Use monospace font for state labels

### LINKED ISSUE #224: State label overflows the state width
In certain case and zoom level the state label overflows the state width. I noticed it depends on the label string. I didn't have a good open source example, so I changed the source code to a hardcoded label `ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklomnopqrstuvxyz` in `time-graph-state.ts` method `renderLabel()`. 

It looks like for strings that start with multiple upper case characters followed lower case characters it happens. For other strings, for example all lower case, all upper case or lower case characters followed by upper case it doesn't happen.

It can be reproduced with the example app in this repository or the theia-trace-extension.

image

image

hoangphamEclipse:
When the timeline chart is zoom in/out, the label of each state will be modified to fit the width of the state. When the width of the state is smaller than the width of the label, the text is truncated. To determine how much text is truncated, the timeline chart uses the ratio of `width of label text / width of state`.  However, since the width of each charater is different, this ratio does not well-represent how much text can be displayed in the label, causing text to overflow out of the state. The solution is to use a 
 `monospace` font, in which all characters are of the same width.

This commit changes the font-family of the labels from `Verdana` to  `monospace`. With this change, the browser will default to the monospace font of the OS and make the texts fit within the states.

Fixes #224.

### DIFF
--- a/timeline-chart/src/time-graph-font-controller.ts
+++ b/timeline-chart/src/time-graph-font-controller.ts
@@ -6,7 +6,7 @@ export class FontController {
     private fontColorMap: Map<number, string>;
     private defaultFontName: string = "LabelFont8White";
 
-    constructor(fontFamily: string = "Verdana") {
+    constructor(fontFamily: string = "monospace") {
         this.fontFamily = fontFamily;
         this.fontNameMap = new Map<number, Map<string, string>>();
         this.fontColorMap = new Map<number, string>();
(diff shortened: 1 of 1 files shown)


bhufmann (approved):
This change fixes the immediate bug. Changing the fontWeight looks better than the original weight. I'm ok to merge it. Your other change #230 can be merged on top it if required.

Issue #227, opened 2022-12-20
Zooming-in in timeline-chart leads too many samples requested from backend


bhufmann:
see https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/895 for detailed description and way to reproduce.

This issue tracker is to track the fix for this regression.

bhufmann:
fixed by #231

PR #231, opened 2023-01-05
Issue #895 Fix

williamsyang-work:
Fixes https://github.com/eclipse-cdt-cloud/theia-trace-extension/issues/895

Previously, world range was not getting updated for fine-resolution updates.  Zooming in was always a fine-resolution update, so the world range stayed the same.  This is why nbTimes was becoming so large while zooming: the entire trace was always being fetched.

It's fine if we update the world range for fine-resolution updates.  I thought this was an optimization, but it doesn't actually do anything (besides bug the code).

World range now updates on fine-resolution calls to maybeFetchNewData and the bug seems to be resolved. 


bhufmann (changes requested):
It looks good to me and fixes the issue. 

Please update the commit message. Give it a short description as first line followed by a more detailed description. 

Thanks a lot for the fix.

bhufmann (approved):
Looks good to me. Fixes the referenced bug. Thanks a lot!

Rodrigoplp-work (approved):

bhufmann:
Fixes #227

PR #232, opened 2023-01-06
Cache label BitMapText objects in a map

hoangphamEclipse:
This commit adds a map to the time-graph-state component as a cache for the state labels. The cache is implemented as a map with the text scaling value (display width / full label width) as the key and the PIXI.BitMapText object as the value. With this change, the timeline-chart can reuse display objects if they were not changed, or if they were created recently.

The size of the cache is fixed, determined by the LABEL_CACHE_SIZE constant. When updating the cache, the timeline-chart will remove the oldest entry in the map. The reasoning behind this is that users are more likely to zoom in/out contiguous zoom levels.

This commit improves the loading time when zooming in and out of the timeline-chart.

### DIFF
--- a/timeline-chart/src/components/time-graph-state.ts
+++ b/timeline-chart/src/components/time-graph-state.ts
@@ -12,13 +12,15 @@ export interface TimeGraphStateStyle {
     borderColor?: number
 }
 
+const LABEL_CACHE_SIZE = 10;
+
 export class TimeGraphStateComponent extends TimeGraphComponent<TimelineChart.TimeGraphState> {
 
     static fontController: FontController = new FontController();
 
     protected _options: TimeGraphStyledRect;
-    private textLabelObject: PIXI.BitmapText | undefined;
     private textWidth: number;
+    private labelCache: Map<number, PIXI.BitmapText> = new Map<number, PIXI.BitmapText>();
 
     constructor(
         id: string,
@@ -55,6 +57,10 @@ export class TimeGraphStateComponent extends TimeGraphComponent<TimelineChart.Ti
         if (!this.model.label) {
             return;
         }
+
+        // Clear the label
+        this.displayObject.removeChildren();
+
         const fontName = TimeGraphStateComponent.fontController.getFontName(this._options.color ? this._options.color : 0, this._options.height - 2) ||
             TimeGraphStateComponent.fontController.getDefaultFontName();
         const position = {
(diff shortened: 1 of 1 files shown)


timeline-chart/src/components/time-graph-state.ts
```diff
@@ -12,13 +12,15 @@ export interface TimeGraphStateStyle {
     borderColor?: number
 }
 
+const LABEL_CACHE_SIZE = 10;
```
MatthewKhouzam, line 15: Too small maybe?

MatthewKhouzam (changes requested):
I tested it, and traced it... it has a 0.1% hit rate on the cache. It needs some tweaking.

Good cache design though.

MatthewKhouzam:
I ran a simple test. 10000 states, all with the same name.

2023-01-11T22:28:07.167Z root INFO Hit rate 0 / 1000
2023-01-11T22:28:08.456Z root INFO Hit rate 0 / 2000
2023-01-11T22:28:08.457Z root INFO Hit rate 0 / 3000
2023-01-11T22:28:09.869Z root INFO Hit rate 0 / 4000
2023-01-11T22:28:09.870Z root INFO Hit rate 0 / 5000
2023-01-11T22:28:12.693Z root INFO Hit rate 0 / 6000
2023-01-11T22:28:12.694Z root INFO Hit rate 0 / 7000
2023-01-11T22:28:13.843Z root INFO Hit rate 0 / 8000
2023-01-11T22:28:18.411Z root INFO Hit rate 0 / 9000
2023-01-11T22:28:19.564Z root INFO Hit rate 0 / 10000
2023-01-11T22:28:21.425Z root INFO Hit rate 0 / 11000
2023-01-11T22:28:22.910Z root INFO Hit rate 0 / 12000
2023-01-11T22:28:27.448Z root INFO Hit rate 0 / 13000
2023-01-11T22:28:28.765Z root INFO Hit rate 0 / 14000
2023-01-11T22:28:31.175Z root INFO Hit rate 0 / 15000
2023-01-11T22:28:32.501Z root INFO Hit rate 0 / 16000
2023-01-11T22:28:36.181Z root INFO Hit rate 0 / 17000
2023-01-11T22:28:37.447Z root INFO Hit rate 0 / 18000
2023-01-11T22:28:38.234Z root INFO Hit rate 0 / 19000
2023-01-11T22:28:40.478Z root INFO Hit rate 0 / 20000

Maybe check the key, that it's the same?

bhufmann (commented):
@hoangphamEclipse @MatthewKhouzam, should the cache be on the chart level instead?

Instances of the TimeGraphStates are re-created when zooming (see `TimeGraphChart.addRow()`). The per state cache yields improvement when calling `TimeGrahChart.updateScaleAndPosition()` when resizing the chart or before fetching new data.

WDYT?

hoangphamEclipse:
Thank you guys, I will take a look at your comment.

hoangphamEclipse:
I tried moving the cache into the chart level with the intention of reusing the `BitMapText` objects for all states that share the same label (so if 5 states have the same label, I would use a single display object). However, each `BitMapText` when displayed has a different coordinate in the canvas, so (to the best of my knowledge) we do need to have separate objects for separate labels (due to the XY-coordinate differences). 

@bhufmann WDYT? Do you have another approach that you have in mind?

bhufmann:
> I tried moving the cache into the chart level with the intention of reusing the `BitMapText` objects for all states that share the same label (so if 5 states have the same label, I would use a single display object). However, each `BitMapText` when displayed has a different coordinate in the canvas, so (to the best of my knowledge) we do need to have separate objects for separate labels (due to the XY-coordinate differences).
> 
> @bhufmann WDYT? Do you have another approach that you have in mind?

That's very true that a text object cannot be shared by states. Text objects with the same text would have to have a pool of objects that would be cached. I'm not clear at the moment if that approach would gain performance improvement, @MatthewKhouzam any suggestions on how to proceed?

PR #234, opened 2023-01-12
Add ellipse and star to the character set of label text

hoangphamEclipse:
This commit adds the ellipse (...) and star unicode characters to the character set of the label text in the time-graph-font controller. This would help display the ellipse symbol in the label text (when the text is truncated) as a single character and takes up the same space as a single character.

### DIFF
--- a/timeline-chart/src/time-graph-font-controller.ts
+++ b/timeline-chart/src/time-graph-font-controller.ts
@@ -27,7 +27,7 @@ export class FontController {
             fill: fontColor === "White" ? "white" : "black",
             fontWeight: "bold"
         };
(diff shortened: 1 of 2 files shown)


bhufmann (approved):
LGTM. Thanks!

timeline-chart/src/time-graph-font-controller.ts
```diff
@@ -72,4 +72,13 @@ export class FontController {
         }
         return fontName ? fontName : "";
     }
+
+    private getCharacterSet(): string[][] {
+        let letters: string[][]= [];
+        letters.push(PIXI.BitmapFont.ASCII[0]);
+        letters.push(['★','★']);
```
MatthewKhouzam, line 79: we don't need the star, I suggested it as a joke to test that it works. But if we want to keep it, let's do that! :)

PR #236, opened 2023-01-20
Grid line color update on theme change

williamsyang-work:
This fixes a bug where the grid line colors were not updating on theme change.  This bug was likely introduced by the viewport implementation.

Grid lines are re-drawn explicitly when the theme changes.


bhufmann (changes requested):
Thanks for this. I tried this change, but it doesn't fix it for. The gridline is still not refreshed for me. Without the the gridlines, the row are not selected after clicking on a row or state. 

I also noticed when zooming quickly multiple times with CTRL + mouse wheel, the gridlines are not always refreshed. As mentioned above, when the gridline is not there the selection is not working. 

Probably, when you fix the gridline refresh on theme change the other case is fixed too. If not we can open a separate tracker then.

bhufmann:
@williamsyang-work any news?

PR #244, opened 2023-04-04
Update ubuntu version and node version in ci-cd.yml

bhufmann:
The version of ubuntu is deprecated. Use 20.04 instead.

Also upgrade node version.

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -17,8 +17,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04]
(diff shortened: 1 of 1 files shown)


bhufmann:
It fixes previous build due to deprecated Ubuntu 18.04. see #245.

marco-miller (approved):
LGTM too.

marco-miller:
> It fixes previous build due to deprecated Ubuntu 18.04. see #245.

Seems like the latter's failing build isn't accessible anymore, but that (this PR) should be ok indeed.

PR #246, opened 2023-04-06
Rescale timeline-chart post zoom using PIXI.Container scale property

hoangphamEclipse:
This commit uses PIXI.Container scale property to rescale the chart after each zoom in/zoom out to instead of the using the manually repositioning each elements in the chart to improve the performance of the timeline chart. The scale of the time-graph-chart is changed everytime the view range changes, and is reset once the server has replied to the front-end.

This commit removes the method updateScaleAndPosition(), which was used to manually rescaling the chart.

There are 3 use cases that are affected by this change:

[1] Zooming in and out using the mouse wheel
[2] Zooming in using the selection tool (right-click and drag)
[3] Zooming in using the action buttons (+/-) on the top right of the chart.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart-cursors.ts
+++ b/timeline-chart/src/layer/time-graph-chart-cursors.ts
@@ -27,6 +27,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     constructor(id: string, protected chartLayer: TimeGraphChart, protected rowController: TimeGraphRowController, style?: { color?: number }) {
         super(id, rowController);
+        this.isScalable = false;
         if (style && style.color) {
             this.color = style.color;
         }
@@ -154,6 +155,7 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
         this.onCanvasEvent('mousedown', this._mouseDownHandler);
         this.stateController.onWorldRender(this._updateHandler);
         this.unitController.onSelectionRangeChange(this._updateHandler);
+        this.stateController.onScaleFactorChange(this._updateHandler);
         this.update();
     }
 
@@ -232,6 +234,10 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     update() {
         if (this.unitController.selectionRange) {
+            /**
+             * When user selects a range on the timeline chart, the selection position must correspond to the cursor of the user,
+             * and not the timeline chart itself since scaling might be applied.
+             */
             const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
             const secondCursorPosition = this.getWorldPixel(this.unitController.selectionRange.end);
             const firstCursorOptions = {
--- a/timeline-chart/src/layer/time-graph-chart-selection-range.ts
+++ b/timeline-chart/src/layer/time-graph-chart-selection-range.ts
@@ -9,6 +9,7 @@ export class TimeGraphChartSelectionRange extends TimeGraphViewportLayer {
 
     constructor(id: string, style?: { color?: number }) {
         super(id);
+        this.isScalable = false;
         if (style && style.color) {
             this.color = style.color;
         }
@@ -34,6 +35,7 @@ export class TimeGraphChartSelectionRange extends TimeGraphViewportLayer {
         this._updateHandler = (): void => this.update();
         this.stateController.onWorldRender(this._updateHandler);
         this.unitController.onSelectionRangeChange(this._updateHandler);
+        this.stateController.onScaleFactorChange(this._updateHandler)
         this.update();
     }
 
@@ -44,6 +46,10 @@ export class TimeGraphChartSelectionRange extends TimeGraphViewportLayer {
 
     update() {
         if (this.unitController.selectionRange) {
+            /**
+             * When user selects a range on the timeline chart, the selection position must correspond to the cursor of the user,
+             * and not the timeline chart itself since scaling might be applied.
+             */
             const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
             const secondCursorPosition = this.getWorldPixel(this.unitController.selectionRange.end);
             if (secondCursorPosition !== firstCursorPosition) {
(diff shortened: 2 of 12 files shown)


timeline-chart/src/time-graph-unit-controller.ts
```diff
@@ -91,7 +95,15 @@ export class TimeGraphUnitController {
         if (this._viewRange.end > this.absoluteRange) {
             this._viewRange.end = this.absoluteRange;
         }
+        const newViewRangeLength = this.viewRangeLength;
+        this._scaleFactor = Number(prevViewRangeLength) / Number(newViewRangeLength);
```
williamsyang-work, line 99: There is a property in the state controller called `zoomFactor` which has change handlers:
$$ZoomFactor = ZF = CanvasDisplayWidth/ViewRangeLength$$
The zoomFactor is updated every change to unitController.viewRange.  Could we move the logic for _scaleFactor into the stateController and move the handlers to onZoomChanged?  It would make more sense to only update the scale when the zoom changes.

Assuming canvasDisplayWidth is interchangeable with absoluteRange, the math works out very similarly:
In unit controller: $$scaleFactor = SF_{uc} = VR_{prev} / VR_{new}$$
In state controller: $$SF_{sc} = ZF_{prev} / ZF_{new} = (CDW_{prev}/VR_{prev}) /(CDW_{new}/VR_{new})$$
Since canvas display width won't change at the same time as view range:
$$CDW_{prev} = CDW_{new} \therefore SF_{sc} = SF_{uc}^{-1} = VR_{new} / VR_{prev}$$

williamsyang-work, line 99: There is a property in the state controller called `zoomFactor` which has change handlers:
$$ZoomFactor = ZF = CanvasDisplayWidth/ViewRangeLength$$
The zoomFactor is updated every change to unitController.viewRange.  Could we move the logic for _scaleFactor into the stateController and move the handlers to onZoomChanged?  It would make more sense to only update the scale when the zoom changes.

Assuming canvasDisplayWidth doesn't change: the math works out very similarly:
In unit controller: $$scaleFactor = SF_{uc} = VR_{prev} / VR_{new}$$
In state controller: $$SF_{sc} = ZF_{prev} / ZF_{new} = (CDW_{prev}/VR_{prev}) /(CDW_{new}/VR_{new})$$
Since canvas display width won't change at the same time as view range:
$$CDW_{prev} = CDW_{new} \therefore SF_{sc} = SF_{uc}^{-1} = VR_{new} / VR_{prev}$$

williamsyang-work: Also, we should probably add a handler to recalculate the zoom factor when the window resizes if it isn't already there.

hoangphamEclipse: Hi William, thank you for your review. I moved everything into the state controller as you have suggested. 

> Could we move the logic for _scaleFactor into the stateController and move the handlers to onZoomChanged?

There are 2 cases right now that the scaleFactor needs to be changed: (1) When the view range changes (which will also corresponds to the zoom change), and (2) when the display canvas width changes (when user resize the timeline chart). Unfortunately, the zoom change event is not raised in the second case, so I cannot put everything in the handler for onZoomChanged (to the best of my knowledge). Please feel free to suggest/correct me if I miss something :).

hoangphamEclipse:
Pushing a fix for a bug that occurs when I close the timeline chart and reopen it, the zooming and panning stop working.

PatrickTasse:
After a zoom in/out, the arrows are temporarily not positioned properly until some refresh fixes them shortly later after a debounce.

Furthermore, sometimes, that late refresh never happens and the arrows are left at the wrong position. I believe this is when zooming out and the existing model with its extra range before and after still covers the new window range and the 'maybeFetchNewData' does not trigger a fetch.

The text labels on states are stretched or squeezed before the refresh, although this is not as critical as the arrow positions. And the text label can also get stuck in that stretched or squeezed state when the refresh does not happen.

hoangphamEclipse:
> After a zoom in/out, the arrows are temporarily not positioned properly until some refresh fixes them shortly later after a debounce.
> 
> Furthermore, sometimes, that late refresh never happens and the arrows are left at the wrong position. I believe this is when zooming out and the existing model with its extra range before and after still covers the new window range and the 'maybeFetchNewData' does not trigger a fetch.

I was able to reproduce this issue. I will fix it.

> The text labels on states are stretched or squeezed before the refresh, although this is not as critical as the arrow positions. And the text label can also get stuck in that stretched or squeezed state when the refresh does not happen.

Unfortunately, it is a drawback of scaling the chart layer: everything in the chart is scaled to the same ratio. I will look into what can be done about it.

hoangphamEclipse:
Added documentation on the fix.

timeline-chart/src/layer/time-graph-chart-cursors.ts
```diff
@@ -232,6 +234,10 @@ export class TimeGraphChartCursors extends TimeGraphChartLayer {
 
     update() {
         if (this.unitController.selectionRange) {
+            /**
+             * When user selects a range on the timeline chart, the selection position must corresponds to the cursor of the user,
```
PatrickTasse, line 238: correspond

timeline-chart/src/layer/time-graph-chart-selection-range.ts
```diff
@@ -44,7 +46,11 @@ export class TimeGraphChartSelectionRange extends TimeGraphViewportLayer {
 
     update() {
         if (this.unitController.selectionRange) {
-            const firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
+            /**
+             * When user selects a range on the timeline chart, the selection position must corresponds to the cursor of the user,
```
PatrickTasse, line 50: correspond

timeline-chart/src/layer/time-graph-chart-selection-range.ts
```diff
@@ -44,7 +46,11 @@ export class TimeGraphChartSelectionRange extends TimeGraphViewportLayer {
...
+             * and not the timeline chart itself since scaling might be applied.
+             */
+            let firstCursorPosition = this.getWorldPixel(this.unitController.selectionRange.start);
```
PatrickTasse, line 53: Looks like it could remain as a const ?

PatrickTasse (changes requested):

PatrickTasse (approved):

PR #250, opened 2023-04-13
[lerna] option verifyAccess to false

marcdumais-work:
We hope this will permit publishing to `npm` using an automation token.

See: https://github.com/lerna/lerna/issues/2788#issuecomment-776726711


bhufmann (approved):
Looks good to me. Thanks for finding this!

PR #258, opened 2023-07-06
Fix time graph glitches when zooming with active filter

PatrickTasse:
- Wait until a new model is applied before setting the new world range and resetting the scale to 1. This could lead to the world range being changed without updating the world model coordinates.

- Change method updateWorldRangeFromViewRange() to computeWorldRangeFromViewRange() that calculates a new world range but does not update the world range. This allows to calculate the world range for a fetch row request without setting the world range yet.

- Do not reset the scale to 1 when a received model is discarded or rejected. This could revert a view range change done by a user operation.

- Do not update world range and scale during full search one-at-a-time row updates. This avoids setting an obsolete world range when a row is returned and accepted after the view range has been changed by a user operation but before this has triggered a new request.

- Add a new field worldZoomFactor to the unit controller. Update the world zoom factor when a new world range is set, but do not update it when the view range is zoomed. Use the world zoom factor to calculate world coordinates in getWorldPixel(), instead of the view range zoom factor.


hoangphamEclipse (approved):
LGTM.

bhufmann (approved):
Great improvement. Thanks for the contribution!

PR #260, opened 2023-07-13
Scale annotations on change of time graph scale factor

PatrickTasse:
- Add the scale factor to the constructor and as a field of a time graph annotation. Use this scale factor when rendering the annotation.

- Rename scaleStateLabels() to scaleComponents() and make the method scale both state labels and annotations. This prevents annotations from being stretched or compressed on zoom.


bhufmann (approved):
Works well. Thanks for the improvement!

PR #276, opened 2024-03-05
Do not package tests

### LINKED ISSUE #275: Big npm package - tests are included
I noticed that the [npm package](https://www.npmjs.com/package/timeline-chart) for `timeline-chart` is way bigger than would be expected:

image

Investigating the reason, it seems tests are included in both source and trans-piled form, including a couple files in the ~25-30MB range. Tests normally are not  needed in npm packages, and should probably not be packaged.
```bash
$ du -hs node_modules/timeline-chart/lib/layer/unitTest/
104M	node_modules/timeline-chart/lib/layer/unitTest/

$ du -hs node_modules/timeline-chart/src/layer/unitTest/
57M	node_modules/timeline-chart/src/layer/unitTest/
```

marcdumais-work:
Fixes #275

This commit adds a `.npmignore` file to help control what's not included in the `timeline-chart` npm package. With it, we now skip packaging the test, which should reduce the package size considerably.

Approx resulting package size (unpacked): from ~160M to ~600k:

image

### DIFF
--- a/timeline-chart/.npmignore
+++ b/timeline-chart/.npmignore
@@ -1,0 +1,4 @@
+unitTest
+__tests__
+jest.config.json
+tsconfig.json


MatthewKhouzam (approved):

marcdumais-work:
@MatthewKhouzam thanks for the review! 

FYI: I had initially tested running `npm pack` in the `timeline-chart` folder, and it seemed to work. But confirming with a local test publish using `verdaccio` and `yarn publish:latest:ci` showed that `.npmignore` needs to be in that sub-folder vs the repo root. So I have made that change, just after your review approval. 

Do you want to have another look?

MatthewKhouzam (approved):
I tested before and saw nothing broke, again, I tested. it seems to work just fine. I did not test on verdaccio, but I can say this breaks nothing from the dev experience. I see the general results too. This is great for our adopters too. So +1.

PR #293, opened 2024-09-10
Throttle fetch requests in time graph chart

PatrickTasse:
If fetching of rows is already in progress, mark any new fetch requests as pending. When the current fetching of rows is complete, executing any pending fetch request.


bhufmann (approved):
Looks good to me. Thanks!

PR #295, opened 2024-09-13
Clamp computed world range to absolute range

PatrickTasse:
Clamp the computed requested world range (which adds a buffer to the left and right of the view range) to the limits of the absolute range.

This prevents unnecessary row updates because the row's provided model range is always clamped.

Reject row data received during full search for an outdated world range.


bhufmann (approved):
Looks good to me.

PR #296, opened 2024-09-19
Hide excluded rows

PatrickTasse:
When rows models are fetched from the data provider, if any requested row is not returned, consider that this row has been excluded by EXCLUDE filtering, and remove its row component. Reposition all remaining rows when rows are excluded.

Keep track of excluded rows and of the world range that was used to exclude them.

During fullSearch with EXCLUDE filter, add previously excluded rows to the request to determine if they should reappear.

When complete update is requested, remove row components that are not visible because they would otherwise not be updated and their states could linger at the wrong position.

After a fullSearch update is completed, restart another request in case new rows have become visible due to space left by newly excluded rows.

Skip coarse update when EXCLUDED filter is applied. This helps prevent matching states that are only found in finer updates from temporarily disappearing during the coarse update.

Adjust the limits when moving vertically considering excluded rows.

### DIFF
--- a/timeline-chart/src/layer/time-graph-chart.ts
+++ b/timeline-chart/src/layer/time-graph-chart.ts
@@ -37,6 +37,7 @@ export type TimeGraphRowStyleHook = (row: TimelineChart.TimeGraphRowModel) => Ti
 
 const VISIBLE_ROW_BUFFER = 3; // number of buffer rows above and below visible range
 const FINE_RESOLUTION_FACTOR = 1; // fine resolution factor or default to disable coarse update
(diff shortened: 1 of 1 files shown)


marcdumais-work:
One dependency is not passing the license check: `electron-to-chromium@1.5.13`

An IP Ticket has already been register against it, earlier today:
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/16197

bhufmann (approved):
Looks good to me. I tested it using https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1120 with small and large traces. Thanks!

PR #338, opened 2026-05-13
Tooltip cursor

MatthewKhouzam:
### What it does

  * Added richer timeline chart cursor behavior with on-chart markers and tooltips.
  * Added support for displaying XY/line-series data within chart rows.
  * Improved chart interaction responsiveness for selection, hover, and pointer actions.
  * Uplift to Pixi 7

### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template

## Summary by CodeRabbit

* **New Features**
  * Added richer timeline chart cursor behavior with on-chart markers and tooltips.
  * Added support for displaying XY/line-series data within chart rows.
  * Improved chart interaction responsiveness for selection, hover, and pointer actions.

* **Bug Fixes**
  * Better handling of cursor and selection visuals so they clear and redraw correctly.
  * Improved chart sizing and background rendering to match the canvas more reliably.

* **Tests**
  * Added coverage for cursor behavior, chart interactions, and XY-series rendering.

### DIFF
--- a/example/webpack.config.js
+++ b/example/webpack.config.js
@@ -19,8 +19,7 @@ module.exports = {
             {
                 enforce: "pre",
                 test: /\.js$/,
-                loader: "source-map-loader",
-                exclude: /node_modules\/@pixi/
+                loader: "source-map-loader"
             },
             {
                 test: /\.png$/,
--- a/example/webpack.config.js
+++ b/example/webpack.config.js
@@ -19,7 +19,8 @@ module.exports = {
             {
                 enforce: "pre",
                 test: /\.js$/,
-                loader: "source-map-loader"
+                loader: "source-map-loader",
+                exclude: /node_modules\/@pixi/
             },
             {
                 test: /\.png$/,
--- a/example/src/index.ts
+++ b/example/src/index.ts
@@ -15,6 +15,7 @@ import { TimeGraphChartGrid } from "timeline-chart/lib/layer/time-graph-chart-gr
 import { TimeGraphVerticalScrollbar } from "timeline-chart/lib/layer/time-graph-vertical-scrollbar";
 import { TimeGraphChartArrows } from "timeline-chart/lib/layer/time-graph-chart-arrows";
 import { TimeGraphRangeEventsLayer } from "timeline-chart/lib/layer/time-graph-range-events-layer";
+import { TimeGraphChartRichCursor } from "timeline-chart/lib/layer/time-graph-chart-rich-cursor";
 
 const styleConfig = {
     mainWidth: 1000,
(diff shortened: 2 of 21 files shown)


MatthewKhouzam:
@jonatanantoni @thorstendb-ARM

JonatanAntoni:
Without having understood the entire code change, I feel like we are mixing two aspects into one here: Cursors and markers. In the view there is this grey mouse marker with the new overlay label and one or two cursors which are sync across all time-based views.

With a mouse click I'd expect a blue cursor is set which does not move along with a mouse move. A second click sets the second blue cursor. Additionally, the grey cursor always moves along with the mouse pointer. Combined with either the shift-modifier or while holding the mouse button down, the range gets selected between the first cursor and the current mouse pointer position.

MatthewKhouzam:
> With a mouse click I'd expect a blue cursor is set which does not move along with a mouse move. A second click sets the second blue cursor. Additionally, the grey cursor always moves along with the mouse pointer. Combined with either the shift-modifier or while holding the mouse button down, the range gets selected between the first cursor and the current mouse pointer position.

The original way of doing it is mouse down == start select, mouse up == end select. The grey follows the mouse as it is a "tooltip cursor"

JonatanAntoni:
> The original way of doing it is mouse down == start select, mouse up == end select. The grey follows the mouse as it is a "tooltip cursor"

Agree, I think we are aligned on the requirements. In my local tests on Mac it just didn't work exactly like that. And I got confused why mouse move events seem to operate on the cursor positions rather then having this entirely separated.

MatthewKhouzam:
> > The original way of doing it is mouse down == start select, mouse up == end select. The grey follows the mouse as it is a "tooltip cursor"
> 
> Agree, I think we are aligned on the requirements. In my local tests on Mac it just didn't work exactly like that. And I got confused why mouse move events seem to operate on the cursor positions rather then having this entirely separated.

please understand I'm in goofing around mode and trying to align with what I understood was a common req too. :)

MatthewKhouzam:
@JonatanAntoni I think this is what you're looking for, try it out.

JonatanAntoni:
@MatthewKhouzam, there is a check failing, not sure about it.
Are you able to reproduce the mouse navigation lags I described above?

What do we need to do to get this merged and wired up into the trace view extension?

cc @thorstendb-ARM

MatthewKhouzam:
Before merging, I want to make sure you like it. ;)

MatthewKhouzam:
Re the CI, it's dash licenses looking like a false positive.

MatthewKhouzam:
About the lags, I feel this was fixed, but please confirm on your side.

example/webpack.config.js
```diff
@@ -19,7 +19,8 @@ module.exports = {
             {
                 enforce: "pre",
                 test: /\.js$/,
-                loader: "source-map-loader"
+                loader: "source-map-loader",
+                exclude: /node_modules\/@pixi/
```
arfio: Why do you exclude pixijs here?

arfio (commented):
Great addition! On my setup I have overlapping tooltip information, making it hard to read, especially in views where we have lots of rows. Would it be possible to either reduce the tooltip height and also making the tooltip position not depend on the xy value ?

Image

MatthewKhouzam:
@arfio  try out the latest one?

JonatanAntoni:
I just retested 53d8e409. It works as expected and I like it.

arfio (dismissed):
Looks good on my end, I hope to see it on the trace extension soon!

MatthewKhouzam:
Thank you so much!

arfio (approved):

MatthewKhouzam:
THanks!

